PR #94, opened 2012-05-29
continuous integration with travis

cholin:
It would be nice if we have continuous integration tests for pygit2 as libgit2 has. This pull requests adds the necessary travis build files to automatically check each commit against our test suite.

The result would be something like http://travis-ci.org/#!/cholin/pygit2/builds/1463817


marvol120-dotcom:
Yes

PR #282, opened 2013-10-31
SSH Public Key authentication

bunjiboys:
This pull request will add support for public key authentication with SSH transports. It adds three new optional arguments to clone_repository, `ssh_pubkey`, `ssh_privkey` and `ssh_passphrase`.

To use public key authentication you would need to supply both the public and the private key, while the passphrase is optional, as passphrases are not required for key pairs.

Example:

``` python
git_clone('ssh://github.com/bunjiboys/pygit2.git', 
          pubkey='/home/user/.ssh/id_rsa.pub', privkey='/home/user/.ssh/id_rsa')
```

Note: there is a bug in the current master branch of libgit2 (has been fixed in development), that prevents the shorter repo urls to be used. In the above example this would cause an "Early EOF" error: `'github.com:bunjiboys/pygit2.git'`. See https://github.com/libgit2/libgit2/pull/1839 for more information.

Also note, that with the current master release of libgit2, if you try to clone using SSH and don't provide valid credentials, you can trigger a segfault (fixed by https://github.com/libgit2/libgit2/commit/219f318c051a37c05dd66687695ebec9f1a90e53)

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -195,9 +195,9 @@ SIMPLE_TYPE(Remote, git_remote, remote)
 
 /* ssh public key auth helper struct */
 typedef struct {
-    char * pubkey;
-    char * privkey;
-    char * passphrase;
+    const char * pubkey;
+    const char * privkey;
+    const char * passphrase;
 } PubKeyAuth;
 
 


remusvrm:
Where can I found the most simple example of cloning a remote repository using SSH credentials?

paqogomez:
@remusavram, quite old, but this page is listed pretty high on google for this problem. 
https://www.pygit2.org/recipes/git-clone-ssh.html

PR #346, opened 2014-02-15
Diff and Patch interface refactored

petrhosek:
This is a complete refactoring of the diff and patch interface. The changes include:
- Splitting `Diff` and `Patch` into separate classes and files, including respective tests. This largely follwos the recent development in `libgit2`.
- Introducing `DiffDelta`, `DiffFile` and `DiffLine` classes and their respective iterators.
- Moving the recently introduced `Blob` diff methods into patch again to be in line with the respective `libgit2` interface.

The biggest difference, apart from the code cleanup, is the lazy evaluation due to heavy use of iterators rather than evaluating everything ahead of time as in case of the existing code, which tends to be slow on larger projects.

I have tried to make the interface as "Pythonic" as possible, while closely following the underlying `libgit2` C interface. However, some of you might think of a better abstraction and I'd happy to incorporate further changes into the patch.


v6:
// , This pull request is stale, no?

jdavid:
Yes, too big of a PR to review. I tried to redo bit by bit and made some commits, but don't have time to finish.

Feel free to propose changes (PRs) concerning Diff if you wish, but avoid huge PR.

jdavid:
Closing, sorry it didn't go through at the time.

PR #405, opened 2014-08-19
Walker interface reimplemented with CFFI

petrhosek:
I've reimplemented the `Walker` interface with CFFI; this is a first step towards reimplementing the entire pygit2 Object model with CFFI. It's supposed to be a straightforward rewrite, I haven't made any changes to the API to avoid breaking the backwards compatibility.


jdavid:
Closing. Nowadays I've a preference for C extension rather than libffi.

PR #415, opened 2014-08-28
Add Repository.ahead_behind()

carlosmn:
This lets us ask how many diverging commits each side of two histories have.

Unfortunately the libgit2 function is currently buggy, so we might want to avoid merging until we can create some tests which are actually accurate.


jdavid:
Hi @carlosmn,

With the move to 0.22, is the libgit2 function still buggy? Will you add some tests?

carlosmn:
It was fixed. I'll see about getting some tests.

carlosmn:
@jdavid test added (not sure if you want more complex, libgit2 itself tests the underlying algo), build green after a little detour with the strarray. Should be ready for merging.

PR #433, opened 2014-10-06
Object or the Index are never False

### LINKED ISSUE #432: Repository.diff fails when treeish is "empty"
Repository.diff contains https://github.com/libgit2/pygit2/blob/master/pygit2/repository.py#L390
This causes a subsequent failure when a is pygit2.Commit and len(a.tree) is 0, because a treeish_to_tree(a) evaluates to False in this case. Consequently, a will be the Commit object afterwards.
An example for such a commit is eb02887bc34130647a52f8152c13531d97f5a058 from https://github.com/twbs/bootstrap.git . https://stackoverflow.com/questions/9765453/gits-semi-secret-empty-tree might shed some light on the empty tree.

carlosmn:
The sequence interface, with `len(X)` returning 0 makes Python consider the object falsy, which is nonsensical. Always consider these to be True.

This fixes #432

### DIFF
--- a/pygit2/index.py
+++ b/pygit2/index.py
@@ -94,6 +94,11 @@ class Index(object):
     def __iter__(self):
         return IndexIterator(self)
 
+    def __bool__(self):
+        return True
+
+    __nonzero__ = __bool__
+
     def read(self, force=True):
         """Update the contents the Index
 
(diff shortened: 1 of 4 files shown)


carlosmn:
The point of this isn't so much about python's behaviour, but our usage of `len()` vs what it seems to expect. We use it for the amount of entries or data content, but it doesn't mean that there is no other information of value there.

We still have the repository, path, whatever else of value. My point here is that an empty file isn't falsy and thus an empty index or blob shouldn't be either.

But if you think that the behaviour is fine, then that's just fine; feel free to close this.

PR #448, opened 2014-11-09
Added installing on OS X

holgi:
This should be considered as a primer to get you started on using pygit2 on OS X. Since I only have one Mac, I could not verify this on more OS Versions and / or settings.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -203,10 +203,32 @@ from a bash shell:
 At this point, you're ready to execute the generic pygit2 installation
 steps described above.
 
+
 Installing on OS X
 ===================================
 
 .. note::
 
-   Help wanted to write this section, please make a pull request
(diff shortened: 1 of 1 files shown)


jdavid:
Hello,

As I understand the PR has not been updated to take into account the @carlosmn comments, @holgi?

I would like to merge this...

holgi:
I had the opportunity to try this on a new Mac with Yosemite installed (since my old one broke down). After installing Xcode and homebrew I could run the following commands without any problems:

```
brew update
brew install libgit2
```

and install pygit2 in a fresh virtualenv with
```
pip3 install pygit2
```

Maybe the stuff to export the libgit2 part could just be left out.

jdavid:
yes I think so, just leaving out that part would be enough

PR #459, opened 2014-12-16
Update to libgit2 v0.22

carlosmn:
The new release should be coming out soon, so let's get ready to update. This new version allows much more customisation of how remotes and repositories are created on clone, which introduced a new concept for us, namely passing ownership of remotes and repositories to the library, which is why there's `_Repository._from_c()` and `_Repository._disown()`.

At the moment there is on feature replacement missing: we no longer expose a boolean to ignore errors but put the handling of certificates in the hands of the programmer. This is done via passing the DER-encoded certificate, to work around the different TLS implementations.

Unfortunately, the parsing done by python's `ssl` module is deep in C code and expects us to have set up a connection via OpenSSL wrapping some socket. But all we can assume is that we're have a buffer with the certificate. If anybody knows of some python module to parse such a buffer, I'd love to know. Everything I've been able to find is either going down to the ASN.1 level or rather incomplete.


aspiers:
It would be awesome if the next release of pygit2 could support the new `describe` stuff introduced by libgit2/libgit2#2592 ;-)

carlosmn:
Sure, but that's rather outside of the scope of this PR.

aspiers:
@carlosmn Oh OK sorry, I guess I misunderstood something.  Why is it outside this PR's scope?

jdavid:
Hi @carlosmn

Now that libgit2 v0.22 has been released, do you think this PR is ready to be merged?

carlosmn:
Yeah, everything that changed should be working with the new version.

PR #461, opened 2014-12-20
Fix use-after-free when patch outlives diff

garious:
And added test that would trigger the bug in Valgrind.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -77,8 +77,8 @@ wrap_patch(git_patch *patch)
 
         delta = git_patch_get_delta(patch);
 
-        py_patch->old_file_path = delta->old_file.path;
-        py_patch->new_file_path = delta->new_file.path;
+        py_patch->old_file_path = strdup(delta->old_file.path);
+        py_patch->new_file_path = strdup(delta->new_file.path);
         py_patch->status = git_diff_status_char(delta->status);
         py_patch->similarity = delta->similarity;
         py_patch->flags = delta->flags;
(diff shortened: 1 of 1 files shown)


garious:
So what do you want to see here?  Boot the test?  Squash the commits to just one that adds `strdup` and `free`?  I'd love to get something merged sooner than later so that I can delete my local fork.  What needs to happen to move forward?

carlosmn:
I'd remove it. Any memory issues need to be looked at with a custom version of python to boot, so I don't see that this running in the normal case brings much.

garious:
Test removed, commits squashed, and rebased on master.

garious:
Thanks!

jdavid:
Hi @garious 

Just noticed that this PR has introduced compilation warnings, could you fix them?

    x86_64-pc-linux-gnu-gcc -pthread -fPIC -I/usr/local/include -I/usr/include/python2.7 -c src/diff.c -o build/temp.linux-x86_64-2.7/src/diff.o
    src/diff.c: In function ‘Patch_dealloc’:
    src/diff.c:156:5: warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [enabled by default]
         free(self->old_file_path);
         ^
    In file included from /usr/include/python2.7/Python.h:42:0,
                     from src/diff.c:29:
    /usr/include/stdlib.h:483:13: note: expected ‘void *’ but argument is of type ‘const char *’
     extern void free (void *__ptr) __THROW;
                 ^
    src/diff.c:157:5: warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [enabled by default]
         free(self->new_file_path);
         ^
    In file included from /usr/include/python2.7/Python.h:42:0,
                 from src/diff.c:29:
    /usr/include/stdlib.h:483:13: note: expected ‘void *’ but argument is of type ‘const char *’
     extern void free (void *__ptr) __THROW;
                 ^

garious:
Fixed: https://github.com/libgit2/pygit2/pull/474

Issue #456, opened 2014-12-02
Unable to push


KINFOO:
Hi,

I am using `pygit2==0.21.4`. I try to push using `KeypairFromAgent` and I am having a Traceback. Here is my code:

``` python
from pygit2 import KeypairFromAgent, Repository

repository = Repository( '/home/kevin/.bee/repositories/bee_test_repository' )
signature = repository.default_signature
credentials = KeypairFromAgent('git')

for remote in repository.remotes:
    remote.credentials = credentials
    remote.push_url = remote.url
    print ('push url: {}'.format(remote.push_url) )
    remote.push('refs/heads/i18n', signature, 'Pushed!')
```

Here comes the Traceback:

``` sh
$ python /tmp/sshkeypairfromagent.py
push url: ssh://git@stash.sierrawireless.local:7999/avtools/bee_test_repository.git
Traceback (most recent call last):
  File "/tmp/sshkeypairfromagent.py", line 11, in <module>
    remote.push('refs/heads/i18n', signature, 'Pushed!')
  File "/home/kevin/.virtualenvs/bee/lib/python3.4/site-packages/pygit2/remote.py", line 358, in push
    check_error(err)
  File "/home/kevin/.virtualenvs/bee/lib/python3.4/site-packages/pygit2/errors.py", line 56, in check_error
    raise GitError(message)
_pygit2.GitError: This transport isn't implemented. Sorry
```

I have read bugs #424 and #379 and [tests](https://github.com/libgit2/pygit2/blob/master/test/test_credentials.py). I still do not get what I am doing wrong. Any clue?

Kevin

Issue #379, opened 2014-06-29
Is there a way to push to repos using the pygit2?


avckp:
When trying to push using ssh I get error.
`ro.push('refs/heads/master')`

`Traceback (most recent call last):`
`File "<input>", line 1, in <module>`
`GitError: This transport isn't implemented. Sorry`

PR #463, opened 2014-12-30
Mentioning libssh2 in remote's pydoc

KINFOO:
Relates to #456

### DIFF
--- a/pygit2/remote.py
+++ b/pygit2/remote.py
@@ -197,7 +197,7 @@ class Remote(object):
     def fetch(self, signature=None, message=None):
         """fetch(signature, message) -> TransferProgress
 
-        Perform a fetch against this remote.
+        Perform a fetch against this remote. May require libssh2.
         """
 
         # Get the default callbacks first
@@ -317,6 +317,7 @@ class Remote(object):
         """push(refspec, signature, message)
 
         Push the given refspec to the remote. Raises ``GitError`` on error.
+        May require libssh2.
 
         :param str spec: push refspec to use
         :param Signature signature: signature to use when updating the tips


carlosmn:
This seems like a really odd place to mention this. It doesn't say why or how it may need libssh2. The referenced issue is about not noticing that the headers for libssh2 weren't installed and as such this text wouldn't have changed the outcome. Adding querying for `git_libgit2_features()` and mentioning that should provide a more direct method by which a user can see if the installation is as they expect.

KINFOO:
It aims to remind users about `libssh2` when they read generated pydoc.

That is what I expected when calls to ssh related methods failed.

@carlosmn You think this has to be removed?

jdavid:
Ideally, when a method requiring libssh2 fails, it should give an informative error to the user. Then this comment in the docstring should be removed.

Looking at issues #379 and #456 today the error is ``GitError: This transport isn't implemented. Sorry``.
So we should catch this error, and give a better message to the user in this case, something like ``ssh transport not supported, please compile libgit2 with libssh2``, or something like that.

Now, looking at the changes in the recently released lbgit2 v0.22, the message ``This transport isn't implemented. Sorry`` is gone. So maybe now libgit2 raises a more informative message by itself, @carlosmn ?

carlosmn:
The error was changed, since the text we had up to 0.21 was the first error message, which was expected to trigger when we had in fact not implemented support for https or ssh. It's been changed to say the url is unuspported.

The message however doesn't mention a particular dependency because that part of libgit2 does not know about the particular dependency. All it knows is whether there is a supported url schema or not, and libssh2 is just the default way to get ssh.

The problem with the mention of libssh2 is that doing the obvious thing (installing libssh2) won't change anything, and it's just one possible dependency. You also need OpenSSL to use https but that's not mentioned. Soon (hopefully) you might also want GnuTLS or SecureTransport to use https. Are we going to mention everything in the hopes the user guesses they need to install the dev package and rebuild libgit2 from that?

Getting the dependencies in the right place is something that needs to happen during the installation of libgit2, which the package manager might handle but if not, these dependencies are something you would want

jdavid:
It could be a generic message, it just needs to give enough clues so the user will find his way. The message up to 0.21 was misleading, it implied an upstream problem (not implemented).

Ok, so lets revert this commit and move to 0.22, we will see if this problem pops again.

jdavid:
commit reverted

PR #468, opened 2015-01-08
Add OS X installation

kxxoling:


jdavid:
Now we have two competing PRs for adding OS X install notes, this one and #448 

Installation instructions should talk just about brew, since installing libgit2 from source or pygit2 with pip is basically like the generic install documentation.

Personally I think #448 is closer. If it has not yet been merged is because I understand it has not been updated to take into account the comments from @carlosmn

Issue #393, opened 2014-07-22
Make docstrings to be nice with pydoc


jdavid:
In some doc-strings we use formatting rules specific to Sphinx, but which make the interactive help (pydoc) to look bad.

This issue is about:
- defining how we should write some recurrent stuff, like documenting input parameters and return values.
- rewrite the doc-strings so they look fine on pydoc, even if that means losing some features with Sphinx, like cross references.

Note. If we had the resources, we would do like Python does, and document things twice: once in the doc-strings and a feature rich version in the rst files. But in any case the doc-strings must be usable with pydoc, so lets start with that.

PR #469, opened 2015-01-09
Remote doc improvements

mduggan:
I was initially a bit confused how to add a refspec to a remote, so I updated the docs to make it clearer.

Fixed some other small doc issues while I was at it - no code changes.

### DIFF
--- a/docs/remotes.rst
+++ b/docs/remotes.rst
@@ -32,6 +32,10 @@ This class contains the data which is available to us during a fetch.
 The Refspec type
 ===================
 
+Refspecs objects are not constructed directly, but returned by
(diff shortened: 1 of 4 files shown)


jdavid:
Regarding the last commit, see issue #393 

Specifically you are reverting commit 5d01300

mduggan:
Ah sorry! I somewhat mindlessly changed that to be more consistent - it seemed to be the only place in the code that wasn't using `:param:`.  If the plan is to dump `:param:` then I'm happy to create a new PR that avoids adding more of them.

jdavid:
Yes that would be cool, we can use the formatting used in commit 5d01300

PR #474, opened 2015-01-14
Fix compiler warnings

garious:
My last patch introduced compiler warnings.  This patch fixes them and one other (`len` was unused).


jdavid:
Thanks!

PR #481, opened 2015-01-29
Add an option to specify root path when creating archives

gandalf013:
Currently, the `write_archive()` creates an archive that extracts the files in the "main root" directory.  This patch allows one to specify a path under which all the files will be extracted.


jdavid:
@carlosmn something else needs to be changed before merging?

carlosmn:
No, it's fine for me. :+1:

Issue #477, opened 2015-01-21
Unit tests fail on i686


bochecha:
I'm trying to update the Fedora Rawhide package ot pygit2 0.22 (we already have libgit2 0.22 there).

On x86_64, unit tests run just fine.

However, on i686, these 3 fail:
```
======================================================================
ERROR: test_push_fast_forward_commits_to_remote_succeeds (test.test_remote.PushTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/pygit2-0.22.0/test/test_remote.py", line 277, in test_push_fast_forward_commits_to_remote_succeeds
    self.remote.push(['refs/heads/master'])
  File "/builddir/build/BUILD/pygit2-0.22.0/build/lib.linux-i686-2.7/pygit2/remote.py", line 337, in push
    check_error(err)
  File "/builddir/build/BUILD/pygit2-0.22.0/build/lib.linux-i686-2.7/pygit2/errors.py", line 41, in check_error
    message = ffi.string(giterr.message).decode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf6 in position 17: ordinal not in range(128)

======================================================================
ERROR: test_push_non_fast_forward_commits_to_remote_fails (test.test_remote.PushTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/pygit2-0.22.0/test/test_remote.py", line 298, in test_push_non_fast_forward_commits_to_remote_fails
    self.assertRaises(pygit2.GitError, self.remote.push, ['refs/heads/master'])
  File "/usr/lib/python2.7/unittest/case.py", line 511, in assertRaises
    callableObj(*args, **kwargs)
  File "/builddir/build/BUILD/pygit2-0.22.0/build/lib.linux-i686-2.7/pygit2/remote.py", line 337, in push
    check_error(err)
  File "/builddir/build/BUILD/pygit2-0.22.0/build/lib.linux-i686-2.7/pygit2/errors.py", line 41, in check_error
    message = ffi.string(giterr.message).decode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 16: ordinal not in range(128)

======================================================================
ERROR: test_push_when_up_to_date_succeeds (test.test_remote.PushTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #487, opened 2015-02-06
Safer handling of string arrays

carlosmn:
We need to keep hold of the strings which we create. We must also hold
on to the array of strings which we assing to our git_strarray.

We were not doing the latter, which meant that our strings may have been
freed too early, leaving us with with memory access errors (though often
not leading to a crash due to the custom allocator in python).

As we need to keep hold of two/three pieces of information, this looks
like a good place to introduce a context manager. This allows us to
keep these pointers alive without burdening the call sites with a return
of multiple objects they have no use for.

### DIFF
--- a/pygit2/index.py
+++ b/pygit2/index.py
@@ -32,7 +32,7 @@ from __future__ import absolute_import, unicode_literals
 from _pygit2 import Oid, Tree, Diff
 from .errors import check_error
 from .ffi import ffi, C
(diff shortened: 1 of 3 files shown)


carlosmn:
@bochecha can you try with this? It should fix #477.

bochecha:
It does. :+1: 

https://koji.fedoraproject.org/koji/taskinfo?taskID=8842754

Thanks a lot.

Issue #458, opened 2014-12-07
TreeEntry equality testing 


ssadler:
So I am comparing 2 different versions of a TreeEntry at the same path in my repository, and the equality test returns True even though they have different OIDs. Just pointing it out, it wasn't hard to find but it seems counter intuitive to me.

    (Pdb) (item1.oid, item2.oid)
    (d21644a6c7570de018ea623d37732b8e549218ec, c9f6f5428e80fb5d3954ccf0ae2ca24c1c999f53)
    (Pdb) (item1.name, item2.name)
    ('tx', 'tx')
    (Pdb) item1 == item2
    True

Otherwise, thanks for the great lib!

carlosmn:
It's a bit unfortunate that we use `git_tree_entry_cmp()` for equality, since that's not what the function is meant to do. It is instead meant for sorting in a git-compatible manner and as such only checks for the name and the filemode, as that's what the sorting needs to do.

jdavid:
I understand this is fixed now PR #488 has been merged.

@ssadler reopen if you still have issues.

PR #488, opened 2015-02-07
Split TreeEntry's equality and sorting

carlosmn:
The value equality for TreeEntry needs to take the id into account. The
function we were using `git_tree_entry_cmp()` is only meant for
git-compatible sorting in a tree and thus does not take that into
account. Change `_richcompare()` to do so.

We then leave the git-copmatible sorting method as `TreeEntry.sortcmp()`
which can be used to ask python to use that as the comparison function
in `sorted()` and friends.

---

It might just be good enough to leave out `TreeEntry.sortcmp()` and just rely on richcompare, since the extra checking only comes in when they're otherwise equal.

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -87,10 +87,20 @@ TreeEntry_oid__get__(TreeEntry *self)
     return TreeEntry_id__get__(self);
 }
 
+static int
+compare_ids(TreeEntry *a, TreeEntry *b)
+{
+    const git_oid *id_a, *id_b;
+    id_a = git_tree_entry_id(a->entry);
+    id_b = git_tree_entry_id(b->entry);
+    return git_oid_cmp(id_a, id_b);
+}
+
 PyObject *
 TreeEntry_richcompare(PyObject *a, PyObject *b, int op)
 {
     PyObject *res;
+    TreeEntry *ta, *tb;
     int cmp;
 
     /* We only support comparing to another tree entry */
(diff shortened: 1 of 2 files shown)


jdavid:
Tests failing with Python 3

carlosmn:
Ah, they removed sort's 'cmp'. I guess I'll just go with just using richcompare.

carlosmn:
I removed `sortcmp()` as we can just use the rich comparison method anyway, and that lets us use `sort()` in the expected way in each version.

jdavid:
Hello, you have fixed the equal and not-equal cases, but it is still using ``git_tree_entry_cmp`` for the other cases. I find it confusing that the same function uses different criteria, either the oid or the name. I would expose ``git_tree_entry_cmp`` through a separate method, what you first did with ``TreeEntry.sortcmp``, still
available at commit 5c792e30c77b71ffa0b9dbb226c9d81297e241af

Then ``TreeEntry.richcompare`` would use only the oid, this is basically a copy of the ``Oid.richcompare``

(Sorry for not reviewing the code before and just stop at the failing test.)

carlosmn:
I don't see what making `TreeEntry.richcompare` equal to `Oid.richcomare` would bring us. I would find it more confusing that way, as the id is just one aspect of the `TreeEntry`. Entries with names "a" and "aa" should go close to each other, regardless of what object they point to.

Two entries would also not be equal just because they point to the same object. The entries for file "a" and "b" are not equal just because the contents of the files they represent happent to be the same.

jdavid:
With this commit we will run into this situation:

    >>> a == b
    False
    >>> a <= b
    True
    >>> a >= b
    True

jdavid:
In other words this is breaking the order relation (the antisymmetry property to be specific).

The issue #458 which spawned this PR says the current behaviour as counter intuitive, or surprising. But breaking the order relation is just as surprising.

This commit is trying to do two different things in the same API. On the other hand I am not certain which would be the best solution. Maybe some use cases could help shed some light: @ssadler ?

carlosmn:
Ah yes, I hadn't thought about the `<=` and `>=` situation. It's unfortunate that sorting and equality are mixed together this way. If `<=` and `>=` is true for two entries, then the equals is true for sorting, but they may or may not be have value equality.

I've updated the commit to always compare the ids when they would sort as equal. As having duplicate entries would be a bug when you're creating a tree (and might thus care about comparing inequality for entries) it doesn't detract from the ability to sort, and it keeps the result consistent, which the previous version did not.

jdavid:
nice solution, sorting by name then by id

PR #491, opened 2015-02-11
Update git-show recipe

rmoehn:
I couldn't get the diff as shown in the git-show recipe. Therefore
update it to what I think it should be. Maybe there is a better way.

Also add a section on how to assemble a git show-like message. It took
me quite some searching in the Python docs to find out how to do it,
especially the date and time part. So this might save people time. I
wanted to add something that gives me a git show --stat equivalent, but
couldn't figure it out.

### DIFF
--- a/docs/recipes/git-show.rst
+++ b/docs/recipes/git-show.rst
@@ -45,7 +45,9 @@ Produce something like a ``git show`` message
 ======================================================================
 
 In order to display time zone information you have to create a subclass
-of tzinfo as described in the `Python datetime documentation`_::
+of tzinfo. In Python 3.2+ you can do this fairly directly. In older
+versions you have to make your own class as described in the `Python
+datetime documentation`_::
 
     from datetime import tzinfo, timedelta
     class FixedOffset(tzinfo):
@@ -67,10 +69,15 @@ of tzinfo as described in the `Python datetime documentation`_::
 
 Then you can make your message:
 
+    >>> # Until Python 2.7.9:
     >>> from __future__ import unicode_literals
     >>> from datetime import datetime
-    >>>
     >>> tzinfo  = FixedOffset(commit.author.offset)
+
+    >>> # From Python 3.2:
+    >>> from datetime import datetime, timezone, timedelta
+    >>> tzinfo  = timezone( timedelta(minutes=commit.author.offset) )
+    >>>
     >>> dt      = datetime.fromtimestamp(float(commit.author.time), tzinfo)
     >>> timestr = dt.strftime('%c %z')
     >>> msg     = '\n'.join(['commit {}'.format(commit.tree_id.hex),
--- a/docs/recipes/git-show.rst
+++ b/docs/recipes/git-show.rst
@@ -44,8 +44,8 @@ Show all files in commit
 Produce something like a ``git show`` message
 ======================================================================
 
-In order to display timezone information you have to create a subclass of
-tzinfo as described in the `Python datetime documentation`_::
+In order to display time zone information you have to create a subclass
+of tzinfo as described in the `Python datetime documentation`_::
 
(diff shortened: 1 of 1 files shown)


docs/recipes/git-show.rst
```diff
@@ -31,7 +31,7 @@ Show SHA hash
 Show diff
 ======================================================================
 
-    >>> diff = commit.tree.diff()
+    >>> diff = commit.parents[0].tree.diff_to_tree(commit.tree)
```
carlosmn: I think we should prefer to recommend `repo.diff(commit.parents[0], commit)` here, as that accepts treeishes in a few forms.

docs/recipes/git-show.rst
```diff
@@ -40,6 +40,22 @@ Show all files in commit
...
 
+======================================================================
+Produce something like a `git show` message
+======================================================================
+
+    >>> from __future__ import unicode_literals
+    >>> from datetime import datetime
+    >>> import pytz
+    >>> tzinfo  = pytz.timezone('Europe/Berlin')
+    >>> dt      = datetime.fromtimestamp(float(commit.commit_time), tzinfo)
```
carlosmn: git-show/git-log use the author time rather than commit time. In order to show what it would show, we'd want to use `commit.author.time` here.

docs/recipes/git-show.rst
```diff
@@ -40,6 +40,22 @@ Show all files in commit
...
     >>>     print(e.name)
 
+======================================================================
+Produce something like a `git show` message
+======================================================================
+
+    >>> from __future__ import unicode_literals
+    >>> from datetime import datetime
+    >>> import pytz
+    >>> tzinfo  = pytz.timezone('Europe/Berlin')
```
carlosmn: The timezone (really offset) is not fixed for the user running git-show, but is read from the commit, in this case from `commit.author.time_offset`.

rmoehn: Have a look at my new commit. I fixed it there already.

docs/recipes/git-show.rst
```diff
@@ -40,6 +40,45 @@ Show all files in commit
...
+            self.__offset = timedelta(minutes = offset)
+
+        def utcoffset(self, dt):
+            return self.__offset
+
+        def tzname(self, dt):
+            return None
+
+        def dst(self, dt):
+            return timedelta(0) # dealing with DST would be too complicated
```
carlosmn: This comment is misleading. We do not handle DST because a fixed-offset timezone doesn't have DST

rmoehn: Well, a fixed-offset timezone isn't really a timezone. At least in my opinion it shouldn't be necessary to wrap a number in an object that is too dumb to return anything else than that number. But that's Python's fault and you're right: the comment is misleading.

carlosmn: Many countries do not observe DST and they thus have in effect a fixed-offset timezone (though since they're political the implementation would never be as simple as this). Personally I would not set a timezone and use something like `"{}{}".format(offset / 60, offset % 60)` to print the offset.

Though I wonder if pygit2 should provide a datetime as part of the signature instead of time and offset separately.

jdavid:
Python 3.2+ has a new ``datetime.timezone`` object. In my opinion it would be better to use it, and just add a comment about Python 2.

docs/recipes/git-show.rst
rmoehn: That would be convenient. The `.format` version would be shorter, but also ugly, since you have to deal with the sign. `123 / 60 == 2`, `-123 / 60 == -3`

rmoehn:
Thanks for pointing that out. I will upload the patch as soon as I've managed to install pygit2 under Python 3.

Issue #436, opened 2014-10-09
Cherry-pick


rixx:
Do you have plans on including libgit2's [cherry-pick](https://libgit2.github.com/libgit2/#HEAD/group/cherrypick) functionality in the near future? 

I'm definitely interested in helping (in case somebody's already working on this) or contributing this feature myself (in case it's not being tackled already).

carlosmn:
Fixed via #492

PR #492, opened 2015-02-12
Add support for cherrypick()

rnicoll:
This adds limited support for the cherrypick() call only (not cherrypick_commit()). Addresses issue #436 

Learning Python, cffi and the internal workings of Git at the moment, would appreciate any feedback.

### DIFF
--- a/test/test_cherrypick.py
+++ b/test/test_cherrypick.py
@@ -1,0 +1,71 @@
+# -*- coding: UTF-8 -*-
+#
+# Copyright 2010-2014 The pygit2 contributors
+#
+# This file is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License, version 2,
+# as published by the Free Software Foundation.
+#
+# In addition to the permissions in the GNU General Public License,
+# the authors give you unlimited permission to link the compiled
+# version of this file into combinations with other programs,
+# and to distribute those combinations without any restriction
+# coming from the use of this file.  (The General Public License
+# restrictions do apply in other respects; for example, they cover
+# modification of the file, and distribution when not linked into
+# a combined executable.)
+#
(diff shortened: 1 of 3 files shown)


test/test_cherrypick.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_cherrypick_remove_conflicts(self):
+        other_branch_tip = '1b2bae55ac95a4be3f8983b86cd579226d0eb247'
+        self.repo.cherrypick(other_branch_tip)
+        idx = self.repo.index
+        conflicts = idx.conflicts
+        self.assertTrue(conflicts is not None)
+        try:
+            conflicts['.gitignore']
+        except KeyError:
+            self.fail("conflicts['.gitignore'] raised KeyError unexpectedly")
```
carlosmn: You don't need to catch this. If an exception is not caught the testing framework will signal it as an error.

rnicoll:
Removed unneeded try-catch and squashed.

jdavid:
Could you please add the new cherrypick method to the documentation (``docs/``) and to the changelog (``README.rst``)?

jdavid:
just updated myself, thanks!

rnicoll:
Cool, was on my to-do list, but.... yeah, juggling projects. Quick question, the .rst files look like they're generated automatically, but I'm not yet familiar enough with Python to know convention here. If they are, can you give me a pointer on where to start reading?

Will come back and get cherrypick-commit later!

jdavid:
No, the ``.rst`` files (together with the docstrings) are the source of the documentation. To build the documentation:

    $ python setup.py build
    $ cd docs/
    $ make html

The documentation is done with http://sphinx-doc.org/, give a look.

PR #499, opened 2015-03-04
Add Submodule type.

pks-t:
Add the submodule type with methods Repository.lookup_submodule(path)->Submodule, Repository.listall_submodules()->[paths], Submodule.open()->Repository and attributes Submodule.path, Submodule.url, Submodule.name.


pks-t:
If the pull request is considered for merge I can provide tests for this. First I'd like some comments though as I'm unexperienced with Python's cffi.

jdavid:
Hello,

This is a feature that could be done from cffi without performance problems. Having a ``Repository.submodules`` like we have a ``Repository.remotes`` may be a friendlier API, what do you think @carlosmn ?

Anyway, the proposed API is fine too. If you add the unit tests, and after a couple of coding review iterations, this will be in a mergeable state.

Basically I leave it up to you whether you want to explore the possibility of ``Respository.submodules``, or just go ahead with this API (pending comments from @carlosmn )

pks-t:
Thanks for your comments. `Repository.submodules` would certainly be a nice way to access submodules. It shouldn't be that much code that's changing, though, so I'll first work on the issues you've highlighted and wait for comments from @carlosmn regarding API changes.

pks-t:
I've fixed the issues you've highlighted and added some basic tests for the preliminary API.

Macsnow14:
may I ask how to **add a submodule** using pygit2?
I tried **clone** a repo which I want to use as a submodule into my repo but it just added a directory not a git submodule.
so... how to add a submodule exactly?

PR #503, opened 2015-03-09
Implement merging of index entries.

pks-t:
This allows us to generate a textual diff of conflicting files in
bare repositories by first performing a merge, retrieving the
IndexEntries of each conflict and then calling
Index.diff_index_entries(...).

### DIFF
--- a/test/test_repository.py
+++ b/test/test_repository.py
@@ -198,6 +198,39 @@ class RepositoryTest(utils.BareRepoTestCase):
         written_sha1 = self.repo.create_blob(data)
         self.assertEqual(hashed_sha1, written_sha1)
 
+    def test_conflicts_in_bare_repository(self):
+        def create_conflict_file(repo, branch, content):
+            oid = repo.create_blob(content)
+            tb = repo.TreeBuilder()
+            tb.insert('conflict', oid, pygit2.GIT_FILEMODE_BLOB)
+            tree = tb.write()
+
+            sig = pygit2.Signature('Author', 'author@example.com')
+            commit = repo.create_commit(branch.name, sig, sig,
+                    'Conflict', tree, [branch.target])
+            self.assertIsNotNone(commit)
+            return commit
+
+        b1 = self.repo.create_branch('b1', self.repo.head.peel())
+        c1 = create_conflict_file(self.repo, b1, 'Conflict 1')
+        b2 = self.repo.create_branch('b2', self.repo.head.peel())
+        c2 = create_conflict_file(self.repo, b2, 'Conflict 2')
+
+        index = self.repo.merge_commits(c1, c2)
+        self.assertIsNotNone(index.conflicts)
+
+        # ConflictCollection does not allow calling len(...) on it directly so
+        # we have to calculate length by iterating over its entries
+        self.assertEqual(sum(1 for _ in index.conflicts), 1)
+
+        (a, t, o) = index.conflicts['conflict']
+        diff = self.repo.merge_file_from_index(a, t, o)
+        self.assertEqual(diff, '''<<<<<<< conflict
+Conflict 1
+=======
+Conflict 2
+>>>>>>> conflict
+''')
 
 class RepositoryTest_II(utils.RepoTestCase):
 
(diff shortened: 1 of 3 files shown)


carlosmn:
The naming here seems odd. This is about merging, rather than diffing. I think this would go better as `Repository.merge_file_from_index()` (or maybe just `merge_file()` which we can overload depending on arguments) since it doesn't particularly need to be a method of `Index` as all we actually need is the repository and the index entries themselves.

We might want to have a convenience method on the index which takes in a path and extracts the three entries and performs the merge, but the basic method should be part of the repo.

pygit2/index.py
```diff
@@ -286,6 +286,27 @@ def diff_to_tree(self, tree, flags=0, context_lines=3, interhunk_lines=0):
...
+        cours = ours._to_c()[0] if ours is not None else ffi.NULL
+        ctheirs = theirs._to_c()[0] if theirs is not None else ffi.NULL
+
+        err = C.git_merge_file_from_index(
+                cmergeresult, self._repo._repo,
+                cancestor, cours, ctheirs,
+                ffi.NULL);
+        check_error(err)
+
+        return ffi.string(cmergeresult.ptr)
```
carlosmn: We need to free the buffer after extracting the string, something like

```python
ret = ffi.string(cmergeresult.ptr)
C.git_buf_free(cmergeresult)

return ret
```

test/test_index.py
```diff
@@ -44,6 +44,31 @@ def test_bare(self):
...
+        b2 = self.repo.create_branch('b2', self.repo.head.peel())
+        c2 = create_conflict_file(self.repo, b2, 'Conflict 2')
+
+        index = self.repo.merge_commits(c1, c2)
+        self.assertIsNotNone(index.conflicts)
+
+        for (a, t, o) in index.conflicts:
+            diff = index.diff_index_entries(a, t, o)
+            self.assertIsNotNone(diff)
+            self.assertGreater(len(diff), 0)
```
carlosmn: Since we know what should come out, I think it'd be better if we compared against a string with the exact expected output, rather than just making sure it's not an empty string.

pks-t:
Agreed. I haven't been comfortable with the naming here, as well. Somehow I had the notion that the index was necessary somewhere, but now I see I've been wrong on that account. Will change that.

pygit2/index.py
pks-t: Fixed now. This also revealed some kind of bug where the ffi.string would get scrambled at the end when not creating it with cmergeresult.len. Seems as if the string is not correctly NULL-terminated in xdiff. I didn't dive into the issue, though.

carlosmn: That's odd, `git_buf` is meant to make sure there's always a NUL past the buffer to allow treating it as a C-string

pks-t: We don't have `git_buf` here, though, but a `git_merge_file_result` where `git_merge_file_result.ptr` is filled by `xdl_merge` in merge_file.c:154.

carlosmn: Oh right, I'm so used to making sure everything's a buf that I didn't notice the `.ptr` wasn't from that. Yeah, this is purely a ptr+len combination rather than a C string.

test/test_repository.py
```diff
@@ -198,6 +198,35 @@ def test_hashfile(self):
...
+
+        b1 = self.repo.create_branch('b1', self.repo.head.peel())
+        c1 = create_conflict_file(self.repo, b1, 'Conflict 1')
+        b2 = self.repo.create_branch('b2', self.repo.head.peel())
+        c2 = create_conflict_file(self.repo, b2, 'Conflict 2')
+
+        index = self.repo.merge_commits(c1, c2)
+        self.assertIsNotNone(index.conflicts)
+
+        for (a, t, o) in index.conflicts:
```
carlosmn: This loop would only ever loop once, and we know which file we want. Possibly asserting `len(index.conflicts)` and then looking up `(a, t, o) = index.conflicts["conflict"]` would show better that we're just testing the one file, rather than saying that every file conflict should be this way.

carlosmn:
Other than the loop in the test, :+1:

jdavid:
argh, forget it, I misread the code :/

PR #513, opened 2015-04-13
Submodule.open() fix, URL write support

pks-t:
Well. As has been initially proposed by @jdavid this PR rewrites the submodule feature I've been implementing in Python with CFFI. Reason has been a problem that I overlooked with the previous implementation, where the repository returned by submodule.open() was not a pygit2.Submodule but of the C SubmoduleType, thus missing some properties/methods. As I found no easy way to properly instantiate the pygit2.Submodule type instead of the C type from C code I just reimplemented that in Python. If there is any obvious way of doing it I may as well do that, if requested.

Some additional functions have been implemented, as well, thus providing URL write support for submodules, their repos and gitmodule entries.


jdavid:
Hello,

I could reproduce the errors reported by travis locally.

Then tested just the first patch, tests pass for Python 2.7 but don't for Python 3

pks-t:
That at least explains why they fail for the Py3 configs. I'll try to fix them later, probably this evening.

pks-t:
Ah, I get what's happening here. The build is tripping over an error in libgit2 which I've already been fixing upstream. As we're not using the newest build the error is obviously still happening. I'll split up the pull request, as the type-fix is applicable without my libgit2 fixes.

PR #519, opened 2015-04-22
remote.py: Denote optional parameters as such

rmoehn:
In the online documentation to Pygit2 it was not visible that (some of)
the parameters to Remote.fetch() and Remote.push() were optional. Fix
this. (I'm not sure if the way I did it is the idiomatic way of marking
a parameter optional in Python docstrings.)

### DIFF
--- a/pygit2/remote.py
+++ b/pygit2/remote.py
@@ -178,7 +178,7 @@ class Remote(object):
         check_error(err)
 
     def fetch(self, signature=None, message=None):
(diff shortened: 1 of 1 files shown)


jdavid:
Actually signatures only need to be defined in C code, it works automatically for Python files. So to reduce development/maintenance costs and improve coherence of the documentation I went through the Python files and removed manually defined signatures from the docstrings.

See commit deb50536f0c3d6dc6fe78fcfd979048e56793674

PR #520, opened 2015-04-23
Keep path references in merge_file_from_index

cjwatson:
IndexEntry._to_c requires its caller to hold a reference to the path it
returns until it no longer needs the C structure.
Repository.merge_file_from_index was not doing so, causing the merge
text to contain garbage from freed memory in some cases.

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -551,9 +551,12 @@ class Repository(_Repository):
         """
         cmergeresult = ffi.new('git_merge_file_result *')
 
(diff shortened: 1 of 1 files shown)


carlosmn:
The single-line expression works fine for the simpler expression, but once we're getting into tuples and negations, it would look a lot more readable if we had a function which had the logic instead of copying on each line, something like

```python
def entry_to_maybe_c(entry):
    if entry is None:
        return ffi.NULL, ffi.NULL

    return entry._to_c()
```

PR #521, opened 2015-04-23
Add hint to Diff.__iter__()

rmoehn:
I'm not the guy who looks at examples in the first place and I guess
there are other people like me. When I wanted find out how to get
information out of a Diff, I looked at the documented methods and didn't
find anything. Only later @cmn showed me the [p for p in diff] example
in the documentation. Add a short piece of information that gives a hint
to those who prefer the dry API docs.


rmoehn:
Sorry for using the wrong username. :grimacing:

PR #525, opened 2015-04-28
Add DiffStats

carlosmn:
This wraps git_diff_stats and can be retrieved through a Diff. It
includes a formatting method.

I wasn't quite sure what to call the formatting function. It's called `_to_buf()` in C but the type is irrelevant in python, and something like `to_string()` didn't seem to fit as it takes options.


carlosmn:
Oh, that python 3, I seem to forget how to create strings each time.

PR #528, opened 2015-05-20
Add support for querying attributes

carlosmn:
Expose a method in the repository which allows querying an attribute for
a file and converts the result to the python equivalent.

---

I'm not completely sure about the name of the method, but the attributes are often called just 'attr' and git has 'git-check-attr'.


jdavid:
For a method or function I prefer to use verbs, it may be ``get_attr``, very much like the libgit2 function it wraps ``git_attr_get``.

carlosmn:
Done.

PR #538, opened 2015-06-19
Add certificate callback.

alexchamberlain:
Existing unit tests pass. Adapted from `__init__.py`.

### DIFF
--- a/test/test_remote.py
+++ b/test/test_remote.py
@@ -242,6 +242,16 @@ class EmptyRepositoryTest(utils.EmptyRepoTestCase):
         self.assertEqual(stats.indexed_objects, REMOTE_REPO_OBJECTS)
         self.assertEqual(stats.received_objects, REMOTE_REPO_OBJECTS)
 
+    def test_fetch_insecure(self):
+        def no_check(certificate, valid, host):
+            return True
+        remote = self.repo.remotes[0]
+        remote.certificate_check = no_check
+        stats = remote.fetch()
+        self.assertEqual(stats.received_bytes, REMOTE_REPO_BYTES)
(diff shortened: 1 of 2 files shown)


test/test_remote.py
```diff
@@ -242,6 +242,16 @@ def test_fetch(self):
         self.assertEqual(stats.indexed_objects, REMOTE_REPO_OBJECTS)
         self.assertEqual(stats.received_objects, REMOTE_REPO_OBJECTS)
 
+    def test_fetch_insecure(self):
+        def no_check(certificate, valid, host):
+            return True
+        remote = self.repo.remotes[0]
+        remote.certificate_check = no_check
```
alexchamberlain: Hmmm this isn't being called, as the remote isn't `https`, but there don't appear to be any `https` tests?

carlosmn: There's a few remotes which do use HTTPS, but this test needs to make sure it's using HTTPS, not assume that some unspecified remote is going to be that way.

alexchamberlain: Sorry; missed this update. Any idea which test case uses HTTPS?

carlosmn: Most should get redirected automatically through GitHub, though some credential tests use it explicitly. That shouldn't matter too much. If you need HTTPS in your test, use a HTTPS url, you're creating a new test, you don't need to use an existing url or remote.

jdavid:
Which is the state of this pull-request?

webknjaz:
@alexchamberlain any updates?

jdavid:
This was implemented in commit 563cb901, just 3 months after this PR was created. Now in commit db3d3990 I've added a unit test.

PR #540, opened 2015-06-26
Update to libgit2 v0.23

carlosmn:
The new release is coming soon, with some large changes in how we deal with remotes, submodules and reflog operations. We don't support enough of submodules for this to be an issue, but the remotes have lost their setters and instead would have to rely on options.


jdavid:
Reviewing the commit for API changes. Looks to me that some things cannot be done now:

- removing remote refspecs, now we can only add
- setting reflog message in ``set_head``
- resetting identity to default behaviour

Also, the documentation is not updated.

carlosmn:
You can remove refspecs via the config. What we don't provide anymore is a false sense of ORM in the remote object. There is a convenience method to add entries as that's a common operation and it's trivially doable. Removal has problems with detecing what the right thing to remove is, as well as it not being how many configuration operations are done with git.

The `set_head()` reflog message was removed as this is a specific operation (switching branches or detaching HEAD) which requires a specific machine-readable message in order for parts of git to behave as expected. Where we can provide these reflog messages, we do.

As for configured ident, if you want to change it only temporarily, you can store the current ident and re-set it after you're done with the impersonation. This kind of behaviour should be really rare.

jdavid:
ok thanks, new version released

PR #559, opened 2015-09-01
(DO NOT MERGE AS IS) Working implementation of Mysql / Mariadb backends + transport

jflesch:
Hello,

I made a working implementation of Mysql / Mariadb backends (ODB + RefDB) + Mysql/Mariadb transport. I assume you could be interested in knowing it exists and that it works quite nicely :).

It relies on a modified version of Libgit2 including Mariadb / Mysql support : https://github.com/jflesch/libgit2/tree/v0.22_mariadb

As of now, it is a little bit slow (I would say, about 3 to 5 times slower than raw files). I'm still trying to fine-tune my MariaDB configuration. However, I guess it's already good-enough to enjoy Mariadb features like Galera replication with Git repositories.

Unfortunately, I'm short on time, so I did all my tests and testing directly from Pygit2 ( https://github.com/jflesch/pygit2/tree/mariadb ). So there are no tests in my Libgit2 branch, so it cannot be officially merged as is :/ . My Git commit history is quite dirty. Initially I did everything in Pygit2 directly until I realized it was a bad idea : I had to make it in Libgit2 directly to have a correct Mysql/Mariadb transport implementation. Additionally, I don't like at all the API I made to instantiate a Repository object relying on Mariadb ; it's ugly :/.

So I guess it can't be merged as is, either in libgit2 or Pygit2, sorry :/


jdavid:
wow :open_mouth:

jflesch:
Woops. fnmatch.c/.h shouldn't be in the branch anymore. My bad. I'll update this PR asap

webknjaz:
@jdavid is there anything related to mysql interface in current master? It would be cool to have the ability to use a read DB backend from Python env...

PR #568, opened 2015-09-26
Move remote callbacks to an interface/object

carlosmn:
Instead of setting the methods/fields in the remote itself, which is a vestige of the older method which libgit2 used, have the user pass in an object with their callbacks, which represents what's going on underneath much more accurately, as well as letting us re-use the callbacks implementation across `Remote` and `clone_repsitory()`.

The docs aren't updated yet, so it's still WIP, but this is what I'd like us to move to.

### DIFF
--- a/docs/remotes.rst
+++ b/docs/remotes.rst
@@ -21,6 +21,12 @@ The Remote type
 .. autoclass:: pygit2.Remote
    :members:
 
+The RemoteCallbacks type
+========================
+
+.. autoclass:: pygit2.RemoteCallbacks
+   :members:
+
 The TransferProgress type
 ===========================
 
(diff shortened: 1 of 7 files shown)


jdavid:
:+1: 

We could support another, less verbose, way to write. Now:

    class MyCallbacks(pygit2.RemoteCallbacks):
        def __init__(self):
            self.credentials = pygit2.UserPass("libgit2", "libgit2")

    callbacks = MyCallbacks()

Or less verbose:

    callbacks = pygit2.RemoteCallbacks(
        credentials=pygit2.UserPass("libgit2", "libgit2"),
    )

carlosmn:
That's a good idea. The credentials or certificate callbacks are some which you sometimes must provide in order to complete the operation, so it makes sense to make them easier to provide in-line.

carlosmn:
This should now give us parity in both. Through the unification we've gained the progress callbacks on clone and the certificate check on fetch and push, instead of having to implement each of them.

PR #572, opened 2015-10-05
tag._message: avoid NULL pointer dereference

olasd:
A tag message can be empty. In that case, git_tag_message returns
NULL. PyBytes_FromString doesn't check its argument for nullness, and
therefore accessing _message on a tag with an empty message segfaults
Python.

### DIFF
--- a/src/tag.c
+++ b/src/tag.c
@@ -111,7 +111,11 @@ PyDoc_STRVAR(Tag__message__doc__, "Tag message (bytes).");
 PyObject *
 Tag__message__get__(Tag *self)
 {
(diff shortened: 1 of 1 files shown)


olasd:
Test case: https://github.com/melo/anyevent--xmpp
```python
>>> import pygit2
>>> repo = pygit2.Repository('.')
>>> repo['a0f7aefc8eb9b70fd01866dae9a62386d71715b3'].message is None
True
>>> repo['a0f7aefc8eb9b70fd01866dae9a62386d71715b3']._message
Erreur de segmentation # (= segfault)
# interpreter exits
```

Issue #573, opened 2015-10-06
Can't clone from HTTP repo requiring authentication


foca:
Not entirely sure if I'm doing something wrong, but I'm running into a wall and I can't figure it out:

```
>>> pygit2.clone_repository('https://github.com/org/repo.git', '/tmp/foo', credentials=pygit2.UserPass('username', 'accessToken'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pygit2/__init__.py", line 293, in clone_repository
    check_error(err)
  File "/usr/local/lib/python2.7/dist-packages/pygit2/errors.py", line 64, in check_error
    raise GitError(message)
_pygit2.GitError: Failed to resolve address for https: Name or service not known
```

This `Failed to resolve address for https: Name or service not known` happens whenever I try to clone or fetch from a repository requiring authentication.

This is the what I'm doing to install pygit2 (under Ubuntu 14.04):

```
sudo apt-get install -y build-essential cmake libssh2-1-dev
sudo apt-get install -y python-dev
sudo apt-get install -y python-pip python-cffi
(
  cd /tmp
  wget https://github.com/libgit2/libgit2/archive/v0.23.2.tar.gz
  tar -zxf v0.23.2.tar.gz
  cd ./libgit2-0.23.2
  cmake .
  make
  sudo make install
)
sudo ldconfig
sudo pip install pygit2
```

carlosmn:
This doesn't touch the authentication. The error is about not finding 'http' as a host. This means that it's trying to resolve the URL as rsync-style ssh.

The most likely situation here is that your libgit2 does not know about https but it does know about ssh. You can check with

```python
import pygit2
bool(pygit2.features & pygit2.GIT_FEATURE_HTTPS)
bool(pygit2.features & pygit2.GIT_FEATURE_SSH)
```

whether the libgit2 used by your installation of pygit2 supports each of those features. If you don't have https, you'll likely need to install the development libraries for OpenSSL for your distribution.

foca:
Installing libssl-dev did it. Thanks.

That makes a lot of sense. The docs fail to mention OpenSSL as a dependency, though. I'll open a PR for that.

Thanks for your time! :heart:

SteffenKockel:
Just for the record: This is still an issue when using salt 3001 on Debian 9 (stretch)

Solved by updating to buster.

PR #574, opened 2015-10-06
List OpenSSL as a dependency in the docs

foca:
For interacting / authenticating against an HTTPS repository we need the OpenSSL development headers present when installing pygit2, yet the documentation fails to mention it, which led to #573.


carlosmn:
Note that OpenSSL is used when we don't have a better alternative. On Windows we go for WinHTTP and on non-ancient OS X we use SecureTransport.

PR #589, opened 2016-01-25
add method to check if a path is ignored

ccope:
This exposes the libgit2 function for checking whether a path is ignored.


jdavid:
Hello! Could you please add unit tests?

PR #593, opened 2016-02-16
Fix repository crash if path passed is not a str

### LINKED ISSUE #588: Repository crashes on bytes paths on Python 3
If the path supplied to `Repository` is a `bytes` objects as in Python 3, pygit2 crashes. This is inconvenient, ~~as all filesystem utilities will return bytes objects~~ they will actually return `str`s. The `os` libraries will accept `bytes` though.

Eg. as used in [powerline][0] raises `must be str, not bytes`

[0]: https://github.com/powerline/powerline/blob/develop/powerline/lib/vcs/git.py#L108

thomwiggers:
This perhaps isnt' the most elegant fix, but otherwise all callers need to be patched.

Closes #588

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -37,6 +37,8 @@ if sys.version_info[0] < 3:
 else:
     from io import BytesIO as StringIO
 
+import six
+
 # Import from pygit2
 from _pygit2 import Repository as _Repository
 from _pygit2 import Oid, GIT_OID_HEXSZ, GIT_OID_MINPREFIXLEN
@@ -56,8 +58,10 @@ from .submodule import Submodule
 
 class Repository(_Repository):
 
-    def __init__(self, *args, **kwargs):
-        super(Repository, self).__init__(*args, **kwargs)
+    def __init__(self, path, *args, **kwargs):
+        if not isinstance(path, six.string_types):
+            path = path.decode('utf-8')
+        super(Repository, self).__init__(path, *args, **kwargs)
         self._common_init()
 
     @classmethod
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ del sys.path[0]
 
 # Python 2 support
 # See https://github.com/libgit2/pygit2/pull/180 for a discussion about this.
+# Using six isn't an option here yet, we don't necessarily have six installed
 if sys.version_info[0] == 2:
     u = lambda s: unicode(s, 'utf-8')
 else:
@@ -186,7 +187,7 @@ extra_args = {
 if cffi_major_version == 0:
     extra_args['ext_modules'].append(ffi.verifier.get_extension())
 else:
-    extra_args['cffi_modules']=['pygit2/_run.py:ffi']
+    extra_args['cffi_modules'] = ['pygit2/_run.py:ffi']
 
 
 setup(name='pygit2',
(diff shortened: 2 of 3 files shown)


thomwiggers:
Ok so that build was a catastrophe, will fix it in a bit.

jdavid:
Hello, could you please provide a unit test? If possible the unit test should be the first commit, and the fix would be the second commit. Extra points for making a new branch on top of current master.

Note too that pygit2 supports both Python 2.7 and 3, and that ``str`` is not the same thing in Python 2 and 3.

Thanks

thomwiggers:
> Note too that pygit2 supports both Python 2.7 and 3, and that `str` is not the same thing in Python 2 and 3.
Yeah, that difference is the whole point. If pygit2 is passed a path represented as `bytes`, it throws a `TypeError`.

I'm actually not sure if this is pygit2's problem or the problem of the applications using pygit2's API. I have the feeling that lots of other parts of pygit2 will also fail if they are passed paths as `bytes`-strings. However, if you reach this decision, it should probably be documented.

(I mainly made this PR and issue #588 because tracking down where the bytes came from in [powerline][1] was more difficult).

[1]: https://github.com/powerline/powerline/

thomwiggers:
I have added a test and rebased on `master`.

pygit2/repository.py
```diff
@@ -56,8 +56,10 @@
 
 class Repository(_Repository):
 
-    def __init__(self, *args, **kwargs):
-        super(Repository, self).__init__(*args, **kwargs)
+    def __init__(self, path, *args, **kwargs):
+        if not isinstance(path, str):
```
pypingou: Shouldn't it be something like ``six.string_types`` here? Because this will surely not work/break on py3

jdavid:
- With Python 2 paths and the like are supposed to be ``str`` (bytes), but the API should accept text (unicode) too.

- With Python 3 paths and the like are supposed to be ``str`` (unicode), but should accept bytes too
(decoded as utf-8).

Basically pygit2 tries to be a good Python 2 citizen and good Python 3 citizen, which develops a kind of
schizophrenia.

At the C level, the relevant line is at ``src/repository.c:100``:

    if (!PyArg_ParseTuple(args, "s", &path))

In Python 2 this is interpreted as bytes or unicode. In Python 3 this only accepts unicode.

jdavid:
I have applied the failing test. Now about the fix.

jdavid:
+1 to @pypingou idea of using six. As the Python side of pygit2 grows six will be useful for this and other
situations.

Then yes the test would be:

    if not isinstance(path, six.string_types):

I leave you, @pypingou or @thomwiggers  some time to add six as dependency :wink: .. otherwise I will do it.

It would be interesting too, but not required, to add tests for paths with non ascii chars.

pygit2/repository.py
thomwiggers: `six` wasn't a dependency of this project so I was hesitant to include it.

It will actually not break though. On py2, you'll pass a `str` or `unicode` to this function. `unicode` has `.decode` and will decode to `str`. py2 `str` is accepted by Repository just fine. On py3 you'll pass a `str` or `bytes` to this function. `bytes` will `.decode` to `str`. py3 `str` is also accepted by `Repository`.

If you pass anything else then it'll break with a vague error about not having `.decode`, but this is the best I could come up with without `six`.

thomwiggers:
FYI: I've looked into fixing it in the C API instead of in the python code, but that does not seem possible because the needed converter is only available in py3.2+: https://docs.python.org/3.5/c-api/unicode.html#c.PyUnicode_FSDecoder

thomwiggers:
I've rebased on master, slightly modified my tests and am now using `six.string_types`.

PR #600, opened 2016-02-28
Allow testing with tox

thomwiggers:
Allow easy testing of the different environments supported by pygit2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,8 @@ class TestCommand(Command):
 
 class sdist_files_from_git(sdist):
     def get_file_list(self):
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -108,7 +108,8 @@ def run(self):
 
 class sdist_files_from_git(sdist):
     def get_file_list(self):
-        popen = Popen(['git', 'ls-files'], stdout=PIPE, stderr=PIPE)
+        popen = Popen(['git', 'ls-files'], stdout=PIPE, stderr=PIPE,
+                      universal_newlines=True)
```
thomwiggers: This also fixes `setup.py sdist` on Python 3.

Issue #616, opened 2016-04-06
pygit2.write_archive() looses executable permissions


ignatenkobrain:
I have repo with some files with executable flag.

Unfortunately, after `pygit2.write_archive()` it looses it.

jdavid:
Should be fixed now PR #898 is merged.

PR #617, opened 2016-04-06
repository: don't loose file mode on write_archive()

ignatenkobrain:
Reference: https://github.com/libgit2/pygit2/issues/616
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -37,6 +37,7 @@ if sys.version_info[0] < 3:
 else:
     from io import BytesIO as StringIO
 
+import os
 import six
 
 # Import from pygit2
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
Unfortunately it's missing test, but I don't have time to write it. Please do it yourself ;)

Thanks!

carlosmn:
There's a couple of issues with this code:

1. It assumes that the current working directory will be the repository working directory's root and it assumes it exists, but more importantly,
2. it stats a file on disk when it's dumping data from an arbitrary tree

The purpose of this method is to write out the contents of the given tree to an archive. The working directory's state should not indicate how we create the result. If the file in the tree _is_ executable and that information is getting lost in the tarball, this still wouldn't be the right solution, because the file in the worktree might not share the exec bit with the file of the same name relative to the current working directory; the current working directory of the script running this may not be the repository's workdir, if it even has one.

PR #634, opened 2016-06-01
repository: add listall_reference_objects() method

seanfarley:
This allows for efficient reading of many references and their targets,
without incurring the overhead of lookup_reference() (which stats for
a loose ref and then reads packed-refs) which can be expensive on NFS
with thousands of refs.

We've been using this patch for a while and I think it makes sense for upstream pygit2. Any feedback on the idea?

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -4,6 +4,7 @@ References
 
 .. contents::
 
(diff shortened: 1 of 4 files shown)


jdavid:
merged, thanks!

(I think I prefer iterators, but that's more work, and ``listall_references`` returns a list anyway)

PR #640, opened 2016-07-06
[RFC] install libgit2 via setup.py

seanfarley:
This is a new fallback for when LIBGIT2 is not defined. If libgit2 is not found, then setup.py will download and build libgit2 so that the correct version matches pygit2.

This is helpful for users on systems that don't have access to install packages or systems with a globally installed libgit that is the incorrect version.

### DIFF
--- a/pygit2/_build.py
+++ b/pygit2/_build.py
@@ -32,7 +32,6 @@ But also used by pygit2 at run time.
 
 # Import from the Standard Library
 import os
-from os import getenv
 
 #
 # The version number of pygit2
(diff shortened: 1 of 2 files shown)


seanfarley:
There's some more cleanup I could do in `setup.py` but before I went down that path, I wanted to bounce this idea off to see if it was even remotely desired.

jdavid:
Hello Sean,

This is out of pygit2's scope in my opinion. An there is already venvgit2 filling the gap:

- https://pypi.python.org/pypi/venvgit2

Maybe we should give visibility to venvgit2 in the pygit2 sites.

Thanks

seanfarley:
Holy moly, I had no idea that project existed! Yeah, it'd be nice to promote venvgit2 for projects that want to deploy the two together.

Issue #655, opened 2016-10-08
Windows tests fail


techtonik:
https://ci.appveyor.com/project/jdavid/pygit2/branch/master/job/pe30m4i7qwl2br64

Windows has a strict requirements that all files need to be closed when removing dir. https://superuser.com/questions/2937/how-do-i-delete-a-folder-thats-in-use

```
======================================================================
ERROR: test_write_commit (test.test_archive.ArchiveTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python35\lib\shutil.py", line 381, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmp0up7s6rm\\testrepo\\.git\\objects\\pack\\pack-e6fbc15b315a0eab6005c44e5b7054b1f0043f39.idx'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "C:\projects\pygit2\test\utils.py", line 135, in tearDown
    self.repo_ctxtmgr.__exit__(None, None, None)
  File "C:\projects\pygit2\test\utils.py", line 86, in __exit__
    rmtree(self.temp_dir)
  File "C:\projects\pygit2\test\utils.py", line 65, in rmtree
    shutil.rmtree(path, onerror=onerror)
  File "C:\Python35\lib\shutil.py", line 488, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Python35\lib\shutil.py", line 378, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python35\lib\shutil.py", line 378, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python35\lib\shutil.py", line 378, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python35\lib\shutil.py", line 378, in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
  File "C:\Python35\lib\shutil.py", line 383, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "C:\projects\pygit2\test\utils.py", line 64, in <lambda>
    onerror = lambda func, path, e: force_rm_handle(func, path, e)
  File "C:\projects\pygit2\test\utils.py", line 47, in force_rm_handle
[...]

PR #656, opened 2016-10-08
Attempt to fix test on Windows

techtonik:
Close index to allow directory removal there (#655)

### DIFF
--- a/pygit2/index.py
+++ b/pygit2/index.py
@@ -47,6 +47,7 @@ class Index(object):
         cindex = ffi.new('git_index **')
         err = C.git_index_open(cindex, to_bytes(path))
         check_error(err)
+        print("index opened")
 
         self._repo = None
         self._index = cindex[0]
@@ -67,6 +68,7 @@ class Index(object):
 
     def __del__(self):
         C.git_index_free(self._index)
+        print("index closed")
 
     def __len__(self):
         return C.git_index_entrycount(self._index)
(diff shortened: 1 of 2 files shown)


techtonik:
@jdavid this needs more serious debugging - https://ci.appveyor.com/project/techtonik/pygit2 - it looks like the problem might be in C level. Index open and closed correctly, and even though it is not clear why it happens 6 times instead of 3, it looks like C level doesn't close the handle. Can you check this with `ptrace` or with `lsof` at the end of archive test?

techtonik:
I can't run `gitless` because of this and run of of time to debug the issue.

carlosmn:
You cannot get this kind of error due to `Index`. The file is closed as soon as the contents are read, and IIRC it's closed even before we start parsing the contents.

What does stay open (and it's what the AppVeyor logs show) are the packfile indices. These are kept open since we memory-map them since they're read-only and are potentially large. In order to close those handles, you need to make sure to free the underlying `git_repository`, which means making sure any `Repository` instances get garbage-collected.

You will then have to work around the object db being made out of read-only files, as Windows won't let you delete them normally, and it's usually a pain to deal with in test suites.

But fixing this in the test suite isn't going to fix anything else using pygit2. Whatever tool you're using must make sure to release its `Repository` instances on Windows if you want to delete repositories while the program is still running.

jdavid:
Thanks for the explanation @carlosmn 

I have merged branch fourplusone/fix/windows-tests from @fourplusone and errors are down to 7 from 146 ... but the build is green, @fourplusone ?

Closing this PR.

PR #667, opened 2016-10-30
Add appveyor testing for python 3.4 and below

fourplusone:
Also fixes garbage collection issue for ConflictCollection


fourplusone:
Appveyor seems to have a problem here:
```Exception calling "UploadFile" with "2" argument(s): "The remote server returned an error: (500) Internal Server Error."```
I don't know if it's an error within the build script or a temporary appveyor issue

PR #670, opened 2016-11-14
Update for v0.25

carlosmn:
We're hoping to have a release candidate for 0.25 this week so there won't be any more API changes.

This PR is just for the changes necessary to keep working with the new version and adding support for the new features like specifying a proxy is out of scope.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -2,7 +2,7 @@
 
 cd ~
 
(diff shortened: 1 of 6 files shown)


carlosmn:
I've updated the code to expect v0.25 as the version now that there's a release out.

Issue #76, opened 2012-03-09
Custom Backends?


tsuraan:
Is there any plan to add custom backends into pygit2?  What Rugged has (Rugged::Backend) looks really nice; it would be great to be able to do that in pygit2, but the entire pygit2 Repository API seems to depend on everything working on a local filesystem.

jmorse:
Hi,

I've taken a look at this matter and prototyped some code linking up libgit2/libgit2-backends with pygit2. It'd be great to get some feedback on whether there's interest in merging it, whether I've missed something etc.

libgit2 neatly separates it's operations from how data is stored through storage backend API's, published in git2/sys/*.h. To use a custom backend with libgit2, one has to build a git_repository around odb / refdb objects configured to use the custom backend rather than the default filesystem backends. Once the git_repository is initialized, the backend is completely transparant to the libgit2 API. Thus, pygit2 would only need to build custom git_repository's to support custom backends.

In the work at jmorse/pygit2@6edb77f5 I've added support for creating pygit2 repo's straight from pre-made git_repository's wrapped in a PyCapsule. I've then put code in a different repo/package (jmorse/pygit2-backends@e1597808) that builds the mysql custom backend, creates or opens a storage backend in a database, constructs a git_repository and wraps it in a pygit2.Repository. Some code to exercise this is in 'testscript.py' in that repo.

Assumptions I've made:
 * Making pygit2 depend on custom backends would be bad, hence the separate repo/package. (I'm no expert on python packaging, perhaps there's a better way).
 * The user of this facility knows that they're not getting a fully fledged git repo out of this. Things like the index, working copy, reflog etc won't work with this technique.

Caveats:
 * I've worked off libgit2 v0.19 and the corresponding pygit2 release. This is the latest release supported for my ubuntu 14.4 machine.
 * Indirect / symbolic references don't work with the mysql backend in it's current state (and probably never will).
 * I've paid no attention to supporting python3
 * No support for custom configuration backends
 * I've had to fix a number of things in the mysql backend, including writing the refdb facility from scratch. Until I get some real tests in place, it may be dodgy. I haven't yet submitted a pull request to the libgit2/libgit2-backends maintainers.

Opinions would be most appreicated.

jdavid:
With PR #942, #948 and #982 merged I consider this done.
Sorry @jmorse I could not look at your work at the time.
Please open new issues if there's something left.

webknjaz:
[...]

PR #690, opened 2017-02-09
Add support for custom backends

hishnash:
We use a `Capsule` to pass the `git_repository` struct through python. 

This `Capsule` does not need a destructor since libgit will free the `git_repository` struct when it needs to.

this addresses some of the needs for https://github.com/libgit2/pygit2/issues/76

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -40,7 +40,7 @@ else:
 import six
 
 # Import from pygit2
-from _pygit2 import Repository as _Repository
+from _pygit2 import Repository as _Repository, init_file_backend
 from _pygit2 import Oid, GIT_OID_HEXSZ, GIT_OID_MINPREFIXLEN
 from _pygit2 import GIT_CHECKOUT_SAFE, GIT_CHECKOUT_RECREATE_MISSING, GIT_DIFF_NORMAL
 from _pygit2 import GIT_FILEMODE_LINK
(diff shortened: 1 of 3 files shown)


pygit2/repository.py
```diff
@@ -915,3 +904,22 @@ def set_ident(self, name, email):
...
+
+class Repository(BaseRepository):
+    def __init__(self, path, *args, **kwargs):
+        if not isinstance(path, six.string_types):
+            path = path.decode('utf-8')
+
+        path_backend = init_file_backend(path)
+        super(Repository, self).__init__(backend=path_backend, *args, **kwargs)
+        self._common_init()
+
```
jdavid: ``_common_init`` would be called twice, first from the super class, and then here

src/pygit2.c
```diff
@@ -144,7 +144,43 @@ hash(PyObject *self, PyObject *args)
...
+PyObject *
+init_file_backend(PyObject *self, PyObject *args)
+{
+    const char* path;
+	int err = GIT_OK;
+	git_repository *repository = NULL;
+	if (!PyArg_ParseTuple(args, "s", &path)) {
+		return NULL;
+	};
+
```
jdavid: could you fix the indentation please? here and elsewhere in the C code

pygit2/repository.py
hishnash: fixed

src/pygit2.c
hishnash: fixed

webknjaz:
Hey @hishnash, is it possible to use this with non-file backends? like mysql or so

hishnash:
@webknjaz yes this was done to allow for https://github.com/luckymarmot/GitPagedArchiveBackend 

here we are just using a custom backed for the ODB portion but you could use something very simlare to get a RefDB backend loaded as well.

webknjaz:
Thank you for the reference, I'll try that :)

PR #692, opened 2017-03-06
Add init_submodules and update_submodules interfaces

jasedit:
This exposes the ability to initialize and update submodules on a repository, which are useful for accessing files in submodules (or being able to use files in a repository which rely on submodule files.)


jasedit:
Odd - the unit tests for submodules all work without error for me on my local machine. I'm uncertain why the unit tests are reporting the construction of the git repositories is failing - if there's any insight as to what may be causing those errors, I'm happy to resolve them.

jasedit:
If I execute the submodule units tests as a standalone unit (e.g. `python -m unittest test/test_submodule.py`) the unit tests pass without error. However, using the command `python setup.py test` causes not only several submodule unit tests to fail (on my machine, these fail now fail with the error reading: `_pygit2.GitError: unexpected data at the end of the pack`), but several other tests suddenly fail, which don't fail on the CI execution.

I've also introduced the flag to allow automatically initializing submodules when executing an update command.

jdavid:
Hi Jason, thanks for contributing.

I don't know about the *unexpected data at the end of the pack* errors, maybe @carlosmn has a clue.

Could you please handle the Windows errors shown in AppVeyor?

jasedit:
Compiling under Visual Studio 10 has been fixed - it turns out the C support under Visual Studio 10 requires all variable declarations at the start of the function, which is different than the other compilers (Is Visual Studio 10 using the C89 standard as opposed to the C99 standard?)

jdavid:
Yes, as far as I know VS is stuck with C89.

jasedit:
Building under Visual Studio 10 works now, though the tests still fail with the mysterious unexpected data error.

jdavid:
I got the same errors locally. So eventually I will look into them, when I find some time.

jasedit:
I asked in the libgit2 Slack for help to make sure I wasn't making calls in an obviously incorrect fashion, but no one there had any major insights into the issue yet. I'm willing to chase down possible leads, but right now I'm at a bit of a loss.

jdavid:
Please rebase over the master branch, the tests should pass now.

Otherwise, cloning pygit2 from github makes the tests very slow; it would be nice to clone from the local disk instead. Also, since this feature is not performance sensible it is good candidate to be implemented with cffi, maybe you want to try.

jasedit:
I had initially attempted to use local repositories for submodules, but ran into errors attempting to operate on the relevant submodules. I don't happen to have the relevant errors on hand, since I resolved them by using a remote submodule.

What were the underlying issues that were interacting with this PR? It looks like there were issues related to running tests in sequence and side effects?

I'm not terribly familiar with writing interfaces using cffi - are there any decent tutorials or examples in the repository I can look at? I've spent a bit of time poking around a few files that appear to use it, but it doesn't seem as obvious as I would hope. I can see some of how to access/create C objects through CFFI, but I have no insight on how to implement callback functions.

jdavid:
The ``test_options`` changed some pygit2 options that made ``test_submodules`` to fail. I fixed it in commit d622e876547 (though now there is 1 test in AppVeyor failing, and I don't have Windows to reproduce/fix it)

A number of features in pygit2 are implemented with cffi already. Anyway this is not a requirement for merging.

PR #696, opened 2017-03-29
Wrap branch_is_checked_out

nhynes:


jdavid:
Hi @nhynes 
This will have to wait for libgit2 v0.26 to be released

Issue #238, opened 2013-05-16
Coding Guidelines


heynemann:
As asked by @jdavid I'd like to start a thread here on the coding guidelines adopted by pygit2 as I got a little confused when submitting my pull request.

Thanks a lot for the help in getting my code merged.

tmr232:
Can this issue be closed? It is very confusing when trying to understand if guidelines are set or not.

esc:
Probably.

tmr232:
@jdavid , can you close the issue?

PR #698, opened 2017-04-05
Better error messages for opening repos.

### LINKED ISSUE #645: error on non-existent Git repository is very non-specific
When I try to open a `Repository` using a path that doesn't actually contain a Git repo, the error is a very non-specific `KeyError`:

``` python
>>> print(pygit2.__version__)
0.24.1
>>> os.mkdir('/tmp/foobarbaz')
>>> gitrepo=pygit2.Repository('/tmp/foobarbaz')

  File "<ipython-input-30-782283f588bb>", line 1, in <module>
    gitrepo=pygit2.Repository('/tmp/foobarbaz')

  File "/usr/local/lib/python2.7/dist-packages/pygit2/repository.py", line 64, in __init__
    super(Repository, self).__init__(path, *args, **kwargs)

KeyError: '/tmp/foobarbaz'
```

It would be helpful to replace this with a more specific `GitError`, so that the error from Git can be distinguished from a "normal" `KeyError` in a statement like this:

``` python
>>> repos = { "repo1":"/path/to/repo1", "repo2":"/path/to/repo2" }
>>> gitrepo = pygit2.Repository( repos["repo3"] )
```

tmr232:
Closes #645.

### DIFF
--- a/src/pygit2.c
+++ b/src/pygit2.c
@@ -168,19 +168,19 @@ init_file_backend(PyObject *self, PyObject *args)
     return PyCapsule_New(repository, "backend", NULL);
 
 cleanup:
-  if (repository) {
-    git_repository_free(repository);
-  }
+    if (repository) {
+        git_repository_free(repository);
+    }
 
-  if (err == GIT_ENOTFOUND) {
-    PyErr_Format(PyExc_Exception,
-            "Repository not found at %s", path);
-  } else {
-    PyErr_Format(PyExc_Exception,
-            "Git error %d while opening repo at %s", err, path);
-  }
-  
-  return NULL;
+    if (err == GIT_ENOTFOUND) {
+        PyErr_Format(PyExc_Exception,
+                "Repository not found at %s", path);
+    } else {
+        PyErr_Format(PyExc_Exception,
+                "Git error %d while opening repo at %s", err, path);
+    }
+
+    return NULL;
 }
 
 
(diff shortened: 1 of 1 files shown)


tmr232:
@jdavid Seems like the same tests fail for every change I submit. Is it something wrong with the tests or the code?

src/pygit2.c
```diff
@@ -151,29 +151,35 @@ PyDoc_STRVAR(init_file_backend__doc__,
...
+  }
+
+  if (err == GIT_ENOTFOUND) {
+    PyErr_Format(PyExc_Exception,
+            "Repository not found at %s", path);
+  } else {
+    PyErr_Format(PyExc_Exception,
+            "Git error %d while opening repo at %s", err, path);
+  }
+
```
jdavid: Thanks, looks good. While you are at it, maybe change style to use 4 spaces indentation. Also, ``Exception`` is too generic, we should use a more specific exception; though choosing a good one is not easy.

tmr232: Isn't there a `GitError` exception we can use?
And, do we have a style guide? I saw #238, but since it's open sinse 2013 I didn't really know what to make of it.

tmr232: Oh, I just realized we have different indentation inside the same function... Fixing that now.

jdavid: To make it simple: PEP 7 for C, PEP 8 for Python.

jdavid: Just pushed a commit to raise GitError for GIT_ENOTFOUND (and to not change the error set by Error_set_str for the other cases, to avoid hiding information).

jdavid:
It should be good for the tests now

PR #700, opened 2017-04-12
Normalize error handling for Repository_lookup_branch

### LINKED ISSUE #681: `lookup_reference` inconsistent return with `lookup_branch`.
`lookup_branch` returns `None` if the branch cannot be found.  `lookup_reference` (and also `revparse_single`) raises a `KeyError` in the same situation.  Returning `None` is the nicer behaviour.

tmr232:
Closes #681

I'm not sure that there isn't other code that needs to be changed around there.

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -32,6 +32,7 @@ from __future__ import absolute_import
 from string import hexdigits
 import sys, tarfile
 from time import time
+
 if sys.version_info[0] < 3:
     from cStringIO import StringIO
 else:
@@ -44,6 +45,7 @@ from _pygit2 import Repository as _Repository, init_file_backend
 from _pygit2 import Oid, GIT_OID_HEXSZ, GIT_OID_MINPREFIXLEN
 from _pygit2 import GIT_CHECKOUT_SAFE, GIT_CHECKOUT_RECREATE_MISSING, GIT_DIFF_NORMAL
 from _pygit2 import GIT_FILEMODE_LINK
+from _pygit2 import GIT_BRANCH_LOCAL, GIT_BRANCH_REMOTE
 from _pygit2 import Reference, Tree, Commit, Blob, Signature
 
 from .config import Config
@@ -57,7 +59,6 @@ from .submodule import Submodule
 
 
 class BaseRepository(_Repository):
-
     def __init__(self, backend, *args, **kwargs):
         super(BaseRepository, self).__init__(backend, *args, **kwargs)
         self._common_init()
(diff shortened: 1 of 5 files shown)


jdavid:
The first commit should be reverted so we don't break backwards compatibility, then the second commit will need to be updated. Maybe do a new single commit to make the PR easier to read.

Iterating over a dict returns the keys not the values, we should do the same.

While we implement the mapping interface, it doesn't need to be identical to a dict. For instance the fact that ``branches.get('')`` raises a ``ValueError`` is correct, since the empty string does not make sense for a branch name.

jdavid:
Sorry, I post some comments to the fist commit, then saw you have addressed them in the second commit, so I have removed those comments.

jdavid:
Maybe the implementation can be simplified if replacing ``local``, ``remote`` and ``top`` by a single option ``flag`` whose values can be ``None``, ``GIT_BRANCH_LOCAL`` or ``GIT_BRANCH_REMOTE``.

If you can make a single commit it will be easier to review.

Thanks!

tmr232:
Fixed the last comment, squashed the commits into one, and also copied the branch test cases & converted them to the new API.
Let's hope it will all be green 😃

jdavid:
Good work, it is almost done 😃 

Could you update the documentation too? There should be no mention of the old API in the documentation.

Thanks!

pygit2/repository.py
```diff
@@ -996,6 +998,54 @@ def set_ident(self, name, email):
...
+            self.local = Branches(repository, flag=GIT_BRANCH_LOCAL)
+            self.remote = Branches(repository, flag=GIT_BRANCH_REMOTE)
+
+    def __getitem__(self, name):
+        branch = None
+        if self._flag & GIT_BRANCH_LOCAL:
+            branch = self._repository.lookup_branch(name, GIT_BRANCH_LOCAL)
+
+        if branch is None and self._flag & GIT_BRANCH_REMOTE:
+            branch = self._repository.lookup_branch(name, GIT_BRANCH_REMOTE)
```
tmr232: This does not work. `lookup_branch` can take `GIT_BRANCH_REMOTE` or nothing.
See:
https://github.com/libgit2/libgit2/blob/master/src/branch.c#L353
https://github.com/libgit2/libgit2/blob/master/src/branch.c#L31

So as far as I can tell, this is needed.

tmr232:
Sure. Should I document in the code as well? Any guidelines for that?

jdavid:
I think the code is clear enough.

tmr232:
I couldn't get the docs to build yet (also, the `os.uname` in `conf.py` is bad for Windows), but I think I got it all.

jdavid:
Merged, thanks!

PR #708, opened 2017-05-07
Randomize OID object hashes

lfos:
Instead of using type punning to convert the OID to a Python hash, use
_Py_HashBytes() to hash the OID again. This means we no longer make any
assumptions on the internal representation of OID values or Python
hashes (before this commit, we at least relied on the fact that OID
hases are longer than Python hashes). Moreover, the random seed stored
in PYTHONHASHSEED is now honored.

This also fixes a compiler warning seen with -Wstrict-aliasing.

Signed-off-by: Lukas Fleischer <lfleischer@lfos.de>

### DIFF
--- a/src/oid.c
+++ b/src/oid.c
@@ -209,8 +209,10 @@ Oid_init(Oid *self, PyObject *args, PyObject *kw)
 Py_hash_t
 Oid_hash(PyObject *oid)
 {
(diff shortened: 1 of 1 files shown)


jdavid:
This doesn't work with Python 2, as it lacks ``_Py_HashBytes``

PR #691, opened 2017-02-27
Worktrees implementation

rmedaer:
I **started** to implement Python binding for worktrees management.
 
Added the following features:
  - Add a new worktree: Repository#add_worktree(name, path) -> None
  - List existing worktrees: Repository#list_worktrees() -> [string]

PR #719, opened 2017-07-28
Rebased worktrees

nhynes:
This is a rebased version of #691 (with a tiny modification to use add+prune opts) if the original author doesn't get around to updating it.


jdavid:
Hi, thanks for the rebase.
Would you like to add some unit tests?

jdavid:
Merged #779

PR #743, opened 2017-11-28
Implement prune remote.fetch

cecedille1:
When fetching a remote, you can use the option prune with a constant to remove branches that do not exist anymore on the remote


jdavid:
Thanks!

Just changed ``assert`` to ``self.assertXXX`` because we are using unittest, not pytest.

PR #745, opened 2017-11-29
Add myself to contributors list 

brandonio21:
I'm not sure if the contributors list is automatically created/updated or if we need to manually do it. I saw some people manually doing it, so I thought that I would too...

I contributed: #741  and #744


jdavid:
Merged.

(The list is updated before every release.)

PR #749, opened 2017-12-01
Fix patch.c windows build errors

brandonio21:
Full disclosure: I don't really understand why Windows builds are [failing for 0.26.2](https://ci.appveyor.com/project/jdavid/pygit2/branch/master). Quick google-foo suggests that it might be because declarations are not first..

I don't have a Windows machine to test on, so I'm going to use Appveyor to do the testing and verification.


brandonio21:
Looks like this fixes the build, but there are still some test failures

jdavid:
Thanks! (The test failures come from a different pull request, will look at them later)

PR #756, opened 2017-12-05
Patch.create_from: Add support for context_lines and interhunk_lines

brandonio21:
Adds support for specifying context_lines and interhunk_lines when calling `Patch.create_from`.

I've also added two tests to test the context lines functionality.

The behavior in this PR mimics the behavior in `tree.c` for adding these flags into the options.


jdavid:
This has introduced a couple of warnings in the tests (Equals vs Equal). Could you fix them?

PR #762, opened 2017-12-22
Add update_submodule method in Repository class.

zhangyiant:


jdavid:
How is this different from ``Repository.update_submodules`` ?
See https://github.com/libgit2/pygit2/blob/master/src/repository.c#L1360 and http://www.pygit2.org/submodule.html?highlight=update_submodules#pygit2.Repository.update_submodules

jdavid:
If the purpose is to type:

    repo.update_submodule(submodule)

Instead of:

    repo.update_submodules([submodule])

I would rather have a single method which accepts both:

    repo.update_submodules(submodule)
    repo.update_submodules([submodule])

Feel free to replace the ``repo.update_submodules`` written in C by one using cffi, which is backwards compatible and accepts a singleton.

Thanks for your PR!

zhangyiant:
The initial intent is to update `submodule` with credential input which is not supported in current `update_submodules` method.
Yes, I agree your suggestion to have a single method. I'll update the PR when it's ready.

zhangyiant:
Close this pull request.
A new pull request will be opened, which will replace current C version update_submodules.

PR #766, opened 2018-02-01
commit: Add support for getting GPG signature data

mark-adams:
This change adds a new `Commit.gpg_signature` property which returns a
tuple containing the GPG signature for a given commit and the bytes that
the signature was computed over. This is accomplished by calling
libgit2's `git_commit_extract_signature()` function.


mark-adams:
Hi @jdavid! I believe this is ready for your review! Thanks! 😄

src/commit.c
```diff
@@ -65,6 +65,39 @@ Commit_message__get__(Commit *commit)
...
+        &gpg_signature, &signed_data, commit->repo->repo, oid, NULL
+    );
+
+    if (err != GIT_OK){
+        git_buf_free(&gpg_signature);
+        git_buf_free(&signed_data);
+
+        if (err == GIT_ENOTFOUND){
+            Py_INCREF(Py_None); Py_INCREF(Py_None);
+            return Py_BuildValue("OO", Py_None, Py_None);
```
jdavid: Don't inc refcount of ``None``, because the **O** already does that.
See https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue for the difference between **O** and **N**

src/commit.c
```diff
@@ -65,6 +65,39 @@ Commit_message__get__(Commit *commit)
...
+
+        return Error_set(err);
+    }
+
+    py_gpg_signature = PyBytes_FromString(gpg_signature.ptr);
+    py_signed_data = PyBytes_FromString(signed_data.ptr);
+    git_buf_free(&gpg_signature);
+    git_buf_free(&signed_data);
+
+    return Py_BuildValue("OO", py_gpg_signature, py_signed_data);
```
jdavid: Here use **NN** instead of **OO**, because ``PyBytes_FromString`` returns a new reference, otherwise there will be a leak.

src/commit.c
mark-adams: Good catch. I haven't developed too many C extensions for Python. Thanks for explaining `O` vs `N`!

src/commit.c
mark-adams: Yep, didn't realize that `OO` did that (same as `O` vs `N` above. 

Fixed!

mark-adams:
@jdavid Thanks for all the feedback. I think I've taken care of all of it if you want to take another look.

PR #771, opened 2018-03-02
git_cred_ssh_key_memory_new() support was added.

ZenSecurity:
Now we can use keys which are stored in memory objects.


jdavid:
Could you add a unit test?

ZenSecurity:
Sure, i will.

PR #774, opened 2018-03-03
Add static Diff.parse_diff

brandonio21:
I think this PR is a little less straight forward than #772 

I think the user should be able to create a Diff object from a unified diff without needing a repository, as in [`git_diff_from_buffer`](https://libgit2.github.com/libgit2/#HEAD/group/diff/git_diff_from_buffer). The existing pygit2 API, however, requires that a Repository be provided when creating a Diff object.

This PR relaxes that requirement and allows `NULL` to be provided for the Repository when creating a Diff object. This allows Diffs to be created without repos.

If this is a good idea, let's merge this in! If not, let's have a discussion. I'm also open to the idea of having an optional kwarg `repository=` so that we can reuse this code in `Repository.parse_diff`

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -831,9 +831,6 @@ Diff_merge(Diff *self, PyObject *args)
     if (!PyArg_ParseTuple(args, "O!", &DiffType, &py_diff))
         return NULL;
 
-    if (py_diff->repo->repo != self->repo->repo)
-        return Error_set(GIT_ERROR);
-
     err = git_diff_merge(self->diff, py_diff->diff);
     if (err < 0)
         return Error_set(err);
--- a/src/diff.c
+++ b/src/diff.c
@@ -54,7 +54,7 @@ wrap_diff(git_diff *diff, Repository *repo)
 
     py_diff = PyObject_New(Diff, &DiffType);
     if (py_diff) {
-        Py_INCREF(repo);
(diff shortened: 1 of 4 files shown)


jdavid:
``Diff_merge`` would need to be updated.

Does it have any sense for the diff object to be attached to the repository? I only found ``Diff_merge`` and it doesn't look to do anything useful with the repo. So I would just remove the ``repo`` from the ``Diff`` type (then ``Repository.parse_diff`` would go away as well).

brandonio21:
To me, it doesn't make much sense. (However, libgit2's `git_diff` struct does indeed have a pointer to a repo).

I think removing the repo from the Diff type is a solid idea, but seems like a very breaking change.

jdavid:
We don't expose the repo, today ``diff.repo`` raises attribute error, so we would not break user code.

But we *may* still need the reference to the repo to keep the libgit2's repo alive.

jdavid:
Well, I think better keep it the way it is, just to be sure the reference is alive.

But then, why to have 2 functions that do essentially the same? Could we just have ``Diff.parse_diff`` and remove ``Repository.parse_diff`` ?

brandonio21:
Oh! I was not aware of the `AttributeError`. In that case, I am happy to remove `Repository.parse_diff`. An update to this PR incoming...

PR #779, opened 2018-03-18
Finish worktree support

terminalmage:
This adds a unit test to the work done in #719 and #691, so that we can get this feature merged. It's also been rebased again, against the head of the `master` branch.


jdavid:
Thanks!

The test is failing in Windows (AppVeyor), could you check that?

terminalmage:
@jdavid This appears to be due to the fact that pygit2 is using forward slashes as path separators, so they fail the comparison. Should be an easy fix using `os.path.realpath()` to normalize it.

terminalmage:
Windows tests passing now. :tada:

jdavid:
Thanks!

Issue #610, opened 2016-03-24
DiffLine.content is represented as unicode


mrh1997:
The field `DiffLine.content` contains a unicode line. As git does not know anything about the encoding of the files to be diffed (they are blobs), I expect this object to be of type `str` in py2 and `bytes` in py3.

Even worse if a file is i.e. latin-1 encoded and contains latin-1 specific characters, all these characters are mapped to '\xfffd'. Thus is impossible to diff non-ascii encoded text files correctly.

I suppose this is a pygit2 bug, as the libgit2.h interface works correctly, as it exposes this field as `const char *` (see https://github.com/libgit2/libgit2/blob/HEAD/include/git2/diff.h#L555)

arcsector:
When you say "The field `DiffLine.content`" are you talking about in the docs or in your file?

Python has weird issues with encodings; I wouldn't be surprised if it was the language itself.

mrh1997:
The [docs](http://www.pygit2.org/diff.html#pygit2.DiffLine.content) do not mention the type at all. But (at least until 0.24.0, which was the last version I tested) the actual implementation of ``DiffLine`` used a variable of type ``unicode`` for the attribute ``content``.

My file is a unicode file. But as git does not know anything about encodings, from git's point of view it is a stream of bytes. Thus I expect ``DiffLine.content`` to be of type bytes (or alternatively that you have to specifiy a encoding somewhere).

What do you mean with "weird issues with encodings"?!? This is definitely a bug in pygit2 (as it uses unicode instead of str/bytes).

arcsector:
1. For me all I had to do was wrap `diffLine.content` with `str()` and everything worked fine. The reason it's in unicode is because git is type agnostic, so if you have a 你 in your code, you don't get busted. I'm not sure why you see this as an issue.

2. I mean system languages don't usually have the issues with encodings that python has. The `errors` flag in python 3.6.5 doesn't work properly like it does in other languages, which means you have to use try/except blocks. But that's out of the scope of this issue.

arcsector:
Make a PR for bytes type, then; you can't convert non-ascii to string on the fly.

jdavid:
This is very much like #790 but concerning ``DiffLine`` instead of ``Patch`` ; a similar solution should apply. Both bytes and text should be available. Discussion on the specific naming is being discussed in #790 , if you want to give your opinion.

Besides ``DiffLine.content`` is a member attribute, I would prefer to use a getter or a method.

[...]

Issue #77, opened 2012-03-13
commit metadata is incorrectly assumed to be UTF-8


bos:
git has not always supported encodings, and so there are commits "in the wild" that contain 8-bit non-UTF-8 messages.

Here's an [example from the git tree itself](https://github.com/gitster/git/commit/c31820c26b8f164433e67d28c403ca0df0316055) that uses ISO-8859-1.

Unfortunately, pygit2 assumes and enforces UTF-8 encodings, in C, at _read_ time. This means that commit metadata that is not UTF-8 causes a `UnicodeDecodeError` to be thrown, in such a way that there is no possible workaround in Python.

As far as I can tell, there are just a few possible sensible approaches:
- Add a `message_raw` field to `Commit` and `Tag` objects that supplies the raw bytes and allows an app to decide how to decode them. This would give a decent fallback. Do the same for author and committer `Signature` objects.
- Perform decoding in `lenient` or `replace` mode. I don't like this because it can lead to silent data loss, which is slightly worse than the current "unrecoverable exception" approach :-)

PR #790, opened 2018-04-27
Patch: Add __str__ and __bytes__ for undecoded content.

erikvanzijst:
`Patch.patch` assumes all content to be encoded in UTF-8 and forcefully
replaces any non-decodable sequences. This can lead to corruption for
content that either does not conform to any specific encoding altogether, or
uses an encoding that is incompatible with, or ambinuous to UTF-8.

This change adds `__str__` and `__bytes__` implementations to `Patch` that
return the unmodified, raw bytes.


erikvanzijst:
Would you mind having a look at this pr, @brandonio21?

erikvanzijst:
We're using PyGit2 to power parts of Bitbucket, but up until now that didn't include patch generation (for which we still just shelled out git). As we're moving more functionality onto pygit, we ran into the forced unicode decoding issue. Bitbucket needs access to the raw, untouched diff/patch bytes and so it seemed appropriate to put that under __str__ and __bytes__.

If I overlooked something and the raw byte contents of patches are already accessible, please let me know.

jdavid:
We should deprecate ``.patch``

So what we need is a way to get bytes and a way to get text:

- ``bytes()``
- ``.decode()``

Since ``.decode()`` will always return text, we can use it to deprecate ``.patch`` ; don't need to implement ``str()``

Don't need to add new tests for ``.decode()``, better to replace ``.patch`` by ``.decode()`` in the tests.

Bonus points if ``.patch`` raises a deprecation warning.

Thanks!

src/patch.c
```diff
@@ -180,30 +180,109 @@ Patch_create_from(PyObject *self, PyObject *args, PyObject *kwds)
...
+    "To access the raw, un-decoded patch, use `str(patch)` (Python 2), or "
+    "`bytes(patch)` (Python 3).");
 
 PyObject *
 Patch_patch__get__(Patch *self)
+{
+    PyErr_WarnEx(PyExc_DeprecationWarning, "`Patch.patch` assumes UTF-8 encoding and can have unexpected results on "
+                                           "other encodings. If decoded text is needed, use `Patch.decode()` "
+                                           "instead. Otherwise use `bytes(Patch)`.", 1);
+    return decode(self, "utf-8", "replace");
```
erikvanzijst: These hardwired parameters are equivalent to the original logic of `to_unicode(buf.ptr, NULL, NULL)`.

jdavid:
It's true that ``str()`` is more obvious than ``.decode()``, but still don't like the idea for ``print(patch)`` to print the text string.

What about ``.as_bytes()`` and ``.as_string()``, or ``.to_bytes()`` and ``.to_string()``? Or something else like that.

Here we're defining the policy that eventually should be used across the whole lib.
For reference, right now only ``Oid`` implements ``str()`` and only ``Blob`` implements ``bytes()``.

erikvanzijst:
> It's true that str() is more obvious than .decode(), but still don't like the idea for print(patch) to print the text string.

That's true only on Python 3 though. What probably shaped my thinking a bit is that most of my day to day work is still Python 2 where `str()` returns raw bytes, not text.

However, I agree that it might be best to put Python 3 idioms ahead of Python 2 and I'm happy to axe `__str__` in that light. Doing that probably also diminishes the value of having a Python 3-specific `__bytes__`.

AFAICT, `Blob` implements the buffer protocol, but `bytes(blob)` only works on Python 3. On Python 2 it relies on `Blob.__str__` which doesn't return the blob's contents. The only 2/3-compatible way to access a blob's bytes seems to be through `Blob.data`, or something like `memoryview(blob).tobytes()` through the buffer protocol interface directly.

What would you think of carrying Blob's interface forward and adding `Patch.data` instead of `as_bytes()`? Additionally I could also implement the buffer protocol onto Patch, so that it behaves identically to Blob.

jdavid:
Objects such as blobs also have ``.read_raw()``, though only blobs have ``.data`` ; not nice to have 2 ways for the same thing. Ideally I would like a consistent way to get bytes and text strings through the library, even if it's not implemented now, I would like it to be defined.

#610 is a similar issue, concerning ``DiffLine``. The names chosen here will apply there. Again we see right now 2 different ways, to get the text string, ``.patch`` and ``.content`` for ``Patch`` and ``DiffLine`` respectively.

So let leave alone ``str``, and ``bytes`` reserved for objects implementing the buffer protocol.

So, ``.data`` looks good enough to me. Here an updated proposal:

- ``.data`` for the bytes string
- ``.text`` for the text string

If somebody wants a text string with an encoding different from utf-8 it would be as easy as:

    .data.decode(...)

So no need for ``.text`` to be a method, it would be just a shorthand.

The buffer protocol I understand is an optimization to reduce memory copies. Whether to implement it or not is up to you, don't know whether it will make a difference for your use case.

These could be the commits:

1. Add ``.data``
2. Add ``.text`` (and deprecate ``.patch``)
3. Implement the buffer protocol

I understand the first one is enough to solve your issue. It's up to you how far do you want to go :smile: 

Please don't hesitate if you have better ideas than ``.data`` and ``.text``

erikvanzijst:
Yeah, I'm happy to add `Patch.data`, as well as a read-only buffer protocol interface. The latter could be helpful when dealing with very large diffs, similar to very large blobs. Buffers probably make less sense for DiffLine.

Though, would `.text` be hardwired to replace invalid code sequences (`errors='replace'`, like Patch.patch does now)? If we do that, making it seemingly trivial to convert between bytes and text, then I worry that we might to continue to feed into the [common misunderstandings](https://github.com/libgit2/pygit2/issues/610#issuecomment-383665869) around text decoding.

I defined `.encode()` to default to `errors='strict'` so as to blow up on non-UTF8 data. A method or attribute that silently truncates its input doesn't seem attractive.

Do we really need a shorthand for decoding a patch/blob/line into text? With Git's lack of encoding information, this is an inherently tricky and unreliable proposition.

jdavid:
This ``replace`` usage was introduced in #77 ;
the commit given as example is https://github.com/gitster/git/commit/c31820c26b8f164433e67d28c403ca0df0316055

That commit doesn't have an ``encoding`` header. Github prefers to print bad data that not printing at all. And so does ``git log`` or ``git show`` for instance.

So I would stick to that behaviour and keep ``replace``. It will also make a smoother transition from ``.patch`` to ``.text``, and through all these years I don't remember anyone raising an issue because of this.

I think it's convenient to have ``.text``, probably we could add it in other places as well, so you can type things like ``print(commit.text)``


For reference there're a few more places where we get the bytes and text with different naming: ``Commit.message``, ``Commit.raw_message``, ``Tag.message``, ``Tag._message``

In some cases there are several attributes available and here we see 2 conventions:

- ``TreeEntry.name`` and ``TreeEntry._name``
- ``Signature.name`` and ``Signature.raw_name``

We need to consider these to have a coherent convention... I like ``Patch.text`` and ``.data`` looks good enough, though it could be ``.raw`` or ``.raw_text``, ??

jnrbsn:
@jdavid @erikvanzijst This can probably be closed now that #893 is merged.

jdavid:
yes, follow up in issue #895 

Thank you both!

Issue #815, opened 2018-09-05
utility to test if a commit is a descendent of one or more commits


chadrik:
Hi,
I'm looking to recreate the behavior of `git branch --contains <commit>` using `pygit2`.   The core of the routine in git is implemented [here](https://github.com/git/git/blob/c05048d43925ab8edcb36663752c2b4541911231/commit.c#L1114) and [here](https://github.com/git/git/blob/c05048d43925ab8edcb36663752c2b4541911231/commit.c#L1132) in `is_descendent_of` and `merge_bases_many`.   I've looked through the `pygit2` source for the building blocks of something similar but couldn't find anything.  

A few questions:
- is there something similar to this in `pygit2` that I missed?
- if no: is this considered too "high level" for this project?
- if no: would you be interested in a pull request for this feature should I find the time to do it?
- if yes: do you think this would be better implemented in C or python?

thanks!

jdavid:
I think there's not. Pull request welcome.

https://libgit2.org/libgit2/#HEAD/group/graph/git_graph_descendant_of and https://libgit2.org/libgit2/#HEAD/group/merge/git_merge_base_many are libgit2's equivalent of those two functions, right?

Then wrap both of them and write the rest in Python.

Thanks!

chadrik:
Sounds easy enough.  Thanks for the guidance.

jdavid:
@chadrik This is done with PR #816 right? Or is there something else to do?

chadrik:
Yup, all done. Thanks for the quick feedback throughout the process.

PR #816, opened 2018-09-06
Add support for git_graph_descendant_of

chadrik:
Fixes issue #815

Let me know if this should be added to the docs, and if so, which section.

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -1133,6 +1133,11 @@ class Branches(object):
     def __init__(self, repository, flag=GIT_BRANCH_ALL, commit=None):
         self._repository = repository
         self._flag = flag
+        if commit is not None:
+            if isinstance(commit, Commit):
+                commit = commit.id
+            elif not isinstance(commit, Oid):
+                commit = self._repository.expand_id(commit)
         self._commit = commit
 
         if flag == GIT_BRANCH_ALL:
@@ -1160,7 +1165,7 @@ class Branches(object):
 
     def __iter__(self):
         for branch_name in self._repository.listall_branches(self._flag):
-            if self._commit is None or self._valid(self[branch_name]):
+            if self._commit is None or self.get(branch_name) is not None:
                 yield branch_name
 
     def create(self, name, commit, force=False):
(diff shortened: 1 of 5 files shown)


chadrik:
Also, do you think it's worth adding a feature like this:

```python
branches = repo.branches.with_commit(oid)
for branch in branches:
    print(branch)
print('master' in branches)
```

I would achieve this with a new `__init__` argument to the `Branches` class that would constrain the results of `__getitem__` and `__iter__` to branches that contained the passed commit id.

jdavid:
Yes please add to the docs, just in `docs/repository.rst`

jdavid:
Yes that sounds good, but would not be better `branches.contains(...)`? to mimic `git branch --contains`

chadrik:
I added the docs.  Let me know if you had something more in mind.

> Yes that sounds good, but would not be better branches.contains(...)? to mimic git branch --contains

That was my original thought, but I felt that `branches.contains()` feels like a method that should return a boolean (do the branches contain something, yes or no?), whereas `branches.with_commit()` seems to better indicate that it will return more branches.  Also, it makes it more clear that it has something to do with commits.  Ultimately, it's your call, though.  Let me know what you think.

pygit2/repository.py
```diff
@@ -1159,14 +1160,23 @@ def get(self, key):
 
     def __iter__(self):
         for branch_name in self._repository.listall_branches(self._flag):
-            yield branch_name
+            if self._valid(self[branch_name]):
```
jdavid: I would rather call `self._valid(branch_name)` to avoid a lookup in the case that `self._commit is None`, which is probably the most common case.

jdavid:
Sounds good, true `with_commit` is more intuitive.

Could you add a test for `with_commit`? See the review as well.

Thanks!

pygit2/repository.py
chadrik: Good point, but I don't want to incur the lookup in the case of `__getitem__` where we already have a branch reference object.   Check my latest commit for what I think is a good compromise.

chadrik:
> Could you add a test for with_commit? See the review as well.

Yep, just wanted to get sign off on the general design first.

chadrik:
Updated!  Let me know what you think.

jdavid:
Merged, thanks!

PR #827, opened 2018-10-17
Add bindings for git_reference_is_valid_name

robinst:
Useful for checking if a branch or tag name is valid, e.g. for a UI that wants to validate a name before trying to create it.

See https://libgit2.org/libgit2/#HEAD/group/reference/git_reference_is_valid_name

I exposed it as a module function `reference_is_valid_name` similar to other free-standing functions that were already there (e.g. `discover_repository`, `hashfile`). A class method on `Reference` would be another option I could look into if you'd prefer that.

### DIFF
--- a/src/pygit2.c
+++ b/src/pygit2.c
@@ -183,6 +183,26 @@ cleanup:
 }
 
 
+PyDoc_STRVAR(reference_is_valid_name__doc__,
+    "reference_is_valid_name(refname) -> bool\n"
+    "\n"
+    "Check if the passed string is a valid reference name.");
+PyObject *
+reference_is_valid_name(PyObject *self, PyObject *args)
+{
+    char* refname;
+    int result;
+
+    if (!PyArg_ParseTuple(args, "es", "utf-8", &refname)) {
+        return NULL;
+    }
+
+    result = git_reference_is_valid_name(refname);
+    PyMem_Free(refname);
+    return PyBool_FromLong(result);
+}
+
+
 PyMethodDef module_methods[] = {
     {"init_file_backend", init_file_backend, METH_VARARGS,
     init_file_backend__doc__},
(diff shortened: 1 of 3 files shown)


src/pygit2.c
```diff
@@ -183,13 +183,34 @@ init_file_backend(PyObject *self, PyObject *args)
...
+    "reference_is_valid_name(refname) -> bool\n"
+    "\n"
+    "Check if the passed string is a valid reference name.");
+PyObject *
+reference_is_valid_name(PyObject *self, PyObject *args)
+{
+    char* refname;
+    int result;
+
+    if (!PyArg_ParseTuple(args, "es", "utf-8", &refname)) {
```
robinst: See [Python's documentation](https://docs.python.org/2/c-api/arg.html) for `"es"`. It allows you to pass in a Unicode string and encodes it as UTF-8.

This is not necessary on Python 3 because `"s"` converts to UTF-8. But in Python 2, it converts to the "default encoding". That makes the test fail like this:

```
>       assert reference_is_valid_name('refs/heads/ünicöde')
E       UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 11: ordinal not in range(128)
```

robinst:
There's an interesting [test failure](https://travis-ci.org/libgit2/pygit2/jobs/442527116) on PyPy:

```python
        with pytest.raises(TypeError):
>           reference_is_valid_name(None)
E           Failed: DID NOT RAISE <type 'exceptions.TypeError'>
```

Should I make that test skip on PyPy?

src/pygit2.c
jdavid: you may use `METH_O` instead of `METH_VARARGS` and then:

    refname = py_str_to_c_str(py_refname, NULL);

that will use utf-8 for Python 2 as well.

robinst:
Thanks, done!

(The AppVeyor build seems to be failing on master too, with the same access denied error.)

PR #850, opened 2018-12-21
Fix "invalid escape sequence \l" warning

stevenwinfield:
Missing raw string, causing a DeprecationWarning in Python 3.6


jdavid:
Applied, thanks!

Issue #852, opened 2018-12-24
git object equality


imbuedHope:
I found that object equality does not work with `pygit2.Commit` objects.

Digging deeper it looks like there is no implementation of `__eq__` (going all the way back to `object`) so it falls back to `__hash__`, and `pygit2.Commit.__hash__` is just falls back to `object.__hash__`.

Which causes this to happen.

```python
python3.6.7
>>> import pygit2
>>> repo = pygit2.Repository('.')
>>> repo.head.get_object() == repo.head.get_object()
False
```

This behavior is confusing and work arounds are quite messy.

imbuedHope:
Did some more poking around and it looks like trying to code this in CPython would fail since `Py_hash_t` is limited to a `long` and even a `short_id` of length 5 wouldn't work. To add to it, python2.7 doesn't also has a cap on the size of `int` which limits the set of `__hash__` entires.

So, basically a solution to this would probably only work for `python3` onward and would have to be monkey pactched onto `_pygit2.Object` the python code. Unless I'm missing something. Not sure if this would be worth merging.

jdavid:
Good catch! PRs welcome.

PR #853, opened 2018-12-26
add hash and equality operations on git objects

imbuedHope:
Resolves issue #852 by implementing `__eq__`, `__ne__`, and `__hash__` for git `_pygit.Object` types. Uses contained `Oid` as a unique identifier and falls back to the `Oid` operations for the same.

The net result is that the following now happens without issue
```python
>>> import pygit2
>>> repo = pygit2.Repository('.')
>>> repo.head.get_object() == repo.head.get_object()
True
>>> repo.head.get_object() != repo.head.get_object()
False
```
Note that unlike `Oid` `__ge__`, `__gt__`, `__le__`, and `__lt__` are not supported. These comparisions don't really make sense for `Object`.

As a bonus `Object` and its subtypes can be used in with `set` and other hash depenent containers. For example,
```python
>>> import pygit2
>>> repo = pygit2.Repository('.')
>>> s = set()
>>> s.add(repo.head.get_object())
>>> s.add(repo.head.get_object())
>>> len(s)
1
```

Note that following is also true, but is probably not a gurantee since a later imlpementation may break it.
```python
>>> import pygit2
>>> repo = pygit2.Repository('.')
>>> hash(repo.head.get_object()) == hash(repo.head.get_object().oid)
True
```

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -181,9 +181,10 @@ Object_peel(Object *self, PyObject *py_type)
 }
 
 Py_hash_t
-Object_hash(PyObject *object)
+Object_hash(Object *object)
 {
-    PyObject *py_oid = PyObject_GetAttrString(object, "oid");
+    const git_oid *oid = git_object_id(object->obj);
+    PyObject *py_oid = git_oid_to_py_str(oid);
     Py_hash_t ret = PyObject_Hash(py_oid);
     Py_DECREF(py_oid);
     return ret;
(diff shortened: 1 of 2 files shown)


jdavid:
Thanks!

It would be more efficient to use `git_object_id` and `git_oid_equal`. Please add unit tests as well.

imbuedHope:
Reimplemented `Object_richcompare` and `Object_hash` to be a bit more efficient and added tests in `test/test_object.py`.

PR #869, opened 2019-02-11
Build manylinux1 OS-specific binary wheels in Travis (CI) and publish them to PyPI (CD)

### LINKED ISSUE #793: Ship wheels for Linux
This is a follow up to #563.

I am trying to pack `gitless` as a snap, and absence of Linux wheel makes the process really painful. It could help if https://github.com/pypa/manylinux wheels were already available.

webknjaz:
Closes #793

In its current state the changed added in this PR enable Travis CI _build and publish_ OS-specific wheel distributions which are compliant with a manylinux1 standard.

This PR has a few intentionally incomplete bits:
1) it requires a PyPI password to be filled in
2) it requires uncommenting deploy stage clause so that publishing would only happen for tagged commits

Considerations
--------------------

For simplicity, this PR reuses image by pyca/cryptography which inherits the official image by pypa/manylinux1 and adds a pre-compiled openssl 1.1.1 on top of it so we don't have to.

As a further build time performance improvement I see how we can improve things by publishing our own image to a public registry (Docker Hub or better quay.io) but it's out of scope of this PR.

Credits
---------

This patch is based on the prior art by @beniwohli and my own experience of building manylinux1 wheels for other projects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 
 cache: pip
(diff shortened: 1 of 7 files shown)


techtonik:
Have you checked the status of `manylinux2010` - it seems to be the way to go https://github.com/pypa/manylinux/issues/179

webknjaz:
Yeah, I know. `pip>19` now supports it. And the latest `auditwheel` has it in the list of platforms.
But first I want to get `manylinux1` done. It's more important since pip on user machines is often outdated and is unable to match `manulinux2020` which would lead to falling back to `sdist` if `manylinux1` is not shipped.

.travis.yml
```diff
@@ -7,22 +8,72 @@ python:
...
 install:
   - pip install .
 
 script:
   - pytest
+
+stages:
+- test
+- name: *deploy_stage
+  if: 1=1 OR tag IS present OR type IN (cron, api)
```
webknjaz: @jdavid
We'll need to apply
```suggestion
  if: tag IS present OR type IN (cron, api)
```
once password is updated.

.travis.yml
```diff
@@ -7,22 +8,72 @@ python:
...
+      # `skip-existing: true` is required to skip uploading dists, already
+      # present in PYPI instead of failing the whole process.
+      # This happens when other CI (AppVeyor etc.) has already uploaded
+      # the very same dist (usually sdist).
+      skip-existing: true
+      user: jdavid
+      password:
+        # Encrypted with `travis encrypt -r aio-libs/aiohttp --api-endpoint 'https://api.travis-ci.com/'`:
+        secure: >-
+          TODO: REPLACE THIS WITH AN ENCRYPTED PASSWORD
```
webknjaz: @jdavid please generate the encrypted string using `travis encrypt -r libgit2/pygit2` command.

And then you can use this page to update in via UI right in this PR: https://github.com/webknjaz/pygit2/edit/feature/manylinux1-binary-wheels/.travis.yml?pr=%2Flibgit2%2Fpygit2%2Fpull%2F869

webknjaz:
@jdavid this is almost ready to be merged.
You need to update encrypted PYPI password and we'll need to update deploy stage appearance condition (which I didn't do for demo purposes).

webknjaz:
_(PR close/reopen trick is just to trigger CI restart)_

PR #872, opened 2019-02-15
Support for new option to skip checking for '.keep' files

hackworks:
This is to support the new option added as part of: https://github.com/libgit2/libgit2/pull/4965


hackworks:
Will rework, need to bump up library version

PR #875, opened 2019-02-19
Build manylinux1 compliant wheel statically

webknjaz:
@jdavid this eliminates shared library dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ jobs:
     python: "3.7"
     services:
     - docker
+    before_install: []
     install: skip
     script: &docker-script
     - travis/build-all-manylinux1-wheels.sh "pygit2" "${LIBGIT2_VERSION}"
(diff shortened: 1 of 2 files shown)


webknjaz:
@techtonik @feisuzhu @beniwohli you would probably be interested in testing this.

@jdavid I still keep having CFFI errors when installing wheel in venv and then importing it from there. But if I `wheel unpack` it and then use `PYTHONPATH=` that dir it imports smoothly and works (except for the TLS cert verification, of course).
Which leads me into thinking that shared object is probably put into non-discoverable location during the build... Ideas?

P.S. This is how this PR looks like in Travis CI in my fork: https://travis-ci.com/webknjaz/pygit2/jobs/178973093

webknjaz:
_(i probably messed up my laptop env somehow: it works fine without a crash on the other machine)_
And yet, it still doesn't explain Travis CI failures...

webknjaz:
Anyway, this PR is good to merge for now. That TLS cert validation failures probably need to be fixed by smth like #538: AFAIU those checks are done within C-code of libgit2, right?

webknjaz:
#876 should be a better way of addressing TLS/HTTPS issue

PR #879, opened 2019-02-24
Allow setting git opt set sll cert locations option in libgit2

### LINKED ISSUE #876: Allow setting GIT_OPT_SET_SSL_CERT_LOCATIONS option in libgit2
Apparently, there's `GIT_OPT_SET_SSL_CERT_LOCATIONS` (ref: https://libgit2.org/libgit2/#HEAD/group/libgit2/git_libgit2_opts) option available in `libgit2` (src: https://github.com/libgit2/libgit2/blob/12c6e1f/src/settings.c#L180-L186).

OpenSSL we have compiled under `manylinux1` image has the following completely non-default address for `OPENSSLDIR` config value:
```console
[root@7a356be85129 ~]# /opt/pyca/cryptography/openssl/bin/openssl version -d
OPENSSLDIR: "/opt/pyca/cryptography/openssl"
```
But even if we wanted to set something more generic we simply wouldn't be able to agree on what such generic value would be. Every OS has it's own "ideal" location in their minds: http://gagravarr.org/writing/openssl-certs/others.shtml

That's why it's important to expose this option so that it could be specified by users who could also have some non-default places for certificates for hardening reasons, for example.

Going forward, Python itself has `ssl` module which might have a better idea of where the current installation stores its certificates. This is the location default of my Gentoo Linux based laptop, for example:
```python
➜ ipython
Python 3.7.1 (default, Jan 28 2019, 08:25:13) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import ssl                                                                                                                                                                             

In [2]: ssl.get_default_verify_paths()                                                                                                                                                         
Out[2]: DefaultVerifyPaths(cafile=None, capath='/etc/ssl/certs', openssl_cafile_env='SSL_CERT_FILE', openssl_cafile='/etc/ssl/cert.pem', openssl_capath_env='SSL_CERT_DIR', openssl_capath='/etc/ssl/certs')
```

So in order to provide better consistency with the runtime, I suggest to not only expose this option but also set the location to match Python's defaults, during initialization.

imbuedHope:
Closes #876 

Adds support for the following

```python3
import pygit2
pygit2.settings.ssl_cert_file = '/path/to/file'
pygit2.settings.ssl_cert_dir = '/path/to/folder'
```
Which is the equivalent of the following in C
```C
git_libgit2_opts(GIT_OPT_SET_SSL_CERT_LOCATIONS, '/path/to/file', NULL);
git_libgit2_opts(GIT_OPT_SET_SSL_CERT_LOCATIONS, NULL, '/path/to/folder');
```

Also adds support for directly loading the settings from the `ssl` module like follows
```python3
import pygit2
import ssl
pygit2.settings._ssl_cert_locations = ssl.get_default_verify_paths()
```

All the new properties that were added to `pygit2.settings.Settings` are write only, since it is write only in `libgit2`.

In addition, `pygit2` also attempts sets `_ssl_cert_locations` based on the `ssl` module on import. This probably won't work on Windows until the wheels are changed to support this in the `libgit2` being shipped with it.

### DIFF
--- a/pygit2/__init__.py
+++ b/pygit2/__init__.py
@@ -28,6 +28,9 @@
 # Import from the future
 from __future__ import absolute_import
 
+# Import from core python modules
+from ssl import get_default_verify_paths as default_ssl_verify_paths
+
 # Low level API
 from _pygit2 import *
 
@@ -267,3 +270,9 @@ def clone_repository(
     return Repository._from_c(crepo[0], owned=True)
 
 settings = Settings()
+
+try:
+    # try to set default ssl cert locations
+    settings._ssl_cert_locations = default_ssl_verify_paths()
+except:
+    pass
--- a/pygit2/settings.py
+++ b/pygit2/settings.py
@@ -33,7 +33,7 @@ from _pygit2 import GIT_OPT_SET_CACHE_OBJECT_LIMIT
 from _pygit2 import GIT_OPT_GET_CACHED_MEMORY
 from _pygit2 import GIT_OPT_ENABLE_CACHING
 from _pygit2 import GIT_OPT_SET_CACHE_MAX_SIZE
-
+from _pygit2 import GIT_OPT_SET_SSL_CERT_LOCATIONS
 
 class SearchPathList(object):
 
@@ -101,4 +101,30 @@ class Settings(object):
         """
         return option(GIT_OPT_SET_CACHE_OBJECT_LIMIT, object_type, value)
 
+    def __set_ssl_cert_file(self, value):
+        """Set the ssl cert file. The value cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, value, None)
+
+    ssl_cert_file = property(fset=__set_ssl_cert_file)
+
+    def __set_ssl_cert_dir(self, value):
+        """Set the ssl cert folder. The value cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, None, value)
+
+    ssl_cert_dir = property(fset=__set_ssl_cert_dir)
+
+    def __set_ssl_cert_locations(self, locations):
+        """
+        Passes both file_path and dir_path to libgit2.
+
+        locations must be a list/tuple type with the the file path
+        at index 0 and the directory path at index 1
+
+        The values set cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, locations[0], locations[1])
+
+    _ssl_cert_locations = property(fset=__set_ssl_cert_locations)
 
--- a/src/options.c
+++ b/src/options.c
@@ -272,6 +272,36 @@ option(PyObject *self, PyObject *args)
             return tup;
         }
 
+        case GIT_OPT_SET_SSL_CERT_LOCATIONS:
+        {
+            PyObject *py_file, *py_dir;
+            const char *file_path, *dir_path;
+            int err;
+
+            py_file = PyTuple_GetItem(args, 1);
+            py_dir = PyTuple_GetItem(args, 2);
+
+            /* py_file and py_dir are only valid if they are strings */
+            if (PyUnicode_Check(py_file) || PyBytes_Check(py_file)) {
+                file_path = py_str_to_c_str(py_file, Py_FileSystemDefaultEncoding);
+            } else {
+                file_path = NULL;
+            }
+
+            if (PyUnicode_Check(py_dir) || PyBytes_Check(py_dir)) {
+                dir_path = py_str_to_c_str(py_dir, Py_FileSystemDefaultEncoding);
+            } else {
+                dir_path = NULL;
+            }
+
+            err = git_libgit2_opts(GIT_OPT_SET_SSL_CERT_LOCATIONS, file_path, dir_path);
+
+            if (err < 0)
+                return Error_set(err);
+
+            Py_RETURN_NONE;
+        }
+
     }
 
     PyErr_SetString(PyExc_ValueError, "unknown/unsupported option value");
--- a/src/pygit2.c
+++ b/src/pygit2.c
@@ -241,6 +241,7 @@ moduleinit(PyObject* m)
     ADD_CONSTANT_INT(m, GIT_OPT_GET_CACHED_MEMORY);
     ADD_CONSTANT_INT(m, GIT_OPT_ENABLE_CACHING);
     ADD_CONSTANT_INT(m, GIT_OPT_SET_CACHE_MAX_SIZE);
+    ADD_CONSTANT_INT(m, GIT_OPT_SET_SSL_CERT_LOCATIONS);
 
     /* Errors */
     GitError = PyErr_NewException("_pygit2.GitError", NULL, NULL);


webknjaz:
Oh, I missed this PR...

P.S. s/sll/ssl/ in title

pygit2/__init__.py
```diff
@@ -28,6 +28,9 @@
 # Import from the future
 from __future__ import absolute_import
 
+# Import from core python modules
+from ssl import get_default_verify_paths as default_ssl_verify_paths
```
webknjaz: Please don't rename this callable. It's an action, not a constant. So I must start with a verb. `get`. Renaming it to constant/var-style is harmful and misleading for readability.
```suggestion
from ssl import get_default_verify_paths
```

pygit2/__init__.py
```diff
@@ -267,3 +270,9 @@ def clone_repository(
     return Repository._from_c(crepo[0], owned=True)
 
 settings = Settings()
+
+try:
+    # try to set default ssl cert locations
+    settings._ssl_cert_locations = default_ssl_verify_paths()
+except:
```
webknjaz: You shouldn't catch all exceptions. Silencing everything makes any related bugs nearly impossible to locate and is harmful therefore. It's important to always fail fast and loudly.
`ssl.get_default_verify_paths()` shouldn't raise any exceptions anyway.
```suggestion
```

pygit2/settings.py
```diff
@@ -101,4 +101,30 @@ def cache_object_limit(self, object_type, value):
         """
         return option(GIT_OPT_SET_CACHE_OBJECT_LIMIT, object_type, value)
 
+    def __set_ssl_cert_file(self, value):
+        """Set the ssl cert file. The value cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, value, None)
+
+    ssl_cert_file = property(fset=__set_ssl_cert_file)
```
webknjaz: It is an ancient way of creating properties. Always use it as a decorator instead.

webknjaz: Also, we should have a getter for this.

pygit2/settings.py
```diff
@@ -101,4 +101,30 @@ def cache_object_limit(self, object_type, value):
         """
         return option(GIT_OPT_SET_CACHE_OBJECT_LIMIT, object_type, value)
 
+    def __set_ssl_cert_file(self, value):
+        """Set the ssl cert file. The value cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, value, None)
```
webknjaz: Setters don't return values.

pygit2/settings.py
```diff
@@ -101,4 +101,30 @@ def cache_object_limit(self, object_type, value):
         """
         return option(GIT_OPT_SET_CACHE_OBJECT_LIMIT, object_type, value)
 
+    def __set_ssl_cert_file(self, value):
```
webknjaz:
```suggestion
    @property
    def ssl_cert_file(self):
        return self._ssl_cert_file

    @ssl_cert_file.setter
    def ssl_cert_file(self, value):
        """Set the ssl cert file."""
        option(GIT_OPT_SET_SSL_CERT_LOCATIONS, value, self.ssl_cert_dir)
        self._ssl_cert_file = value
```

pygit2/settings.py
webknjaz: When setting one value we shouldn't reset another. If a user wants to do this they can update it to `None` explicitly.

pygit2/settings.py
```diff
@@ -101,4 +101,30 @@ def cache_object_limit(self, object_type, value):
...
+    ssl_cert_file = property(fset=__set_ssl_cert_file)
+
+    def __set_ssl_cert_dir(self, value):
+        """Set the ssl cert folder. The value cannot be read.
+        """
+        return option(GIT_OPT_SET_SSL_CERT_LOCATIONS, None, value)
+
+    ssl_cert_dir = property(fset=__set_ssl_cert_dir)
+
+    def __set_ssl_cert_locations(self, locations):
```
webknjaz: This one must be a method. Using property here is odd.

imbuedHope: I went with properties because all other supported settings are implemented as property. It does not make sense to break consistency.

webknjaz: As I wrote earlier, we should go for Pythonic principles. And if they are not respected in some places it doesn't mean that you should follow the lead. When you write in C you don't follow Assembly code-style just because it's an underlying somewhere, do you?

pygit2/settings.py
imbuedHope: There is no get support for the ssl cert paths in `libgit2`. It makes sense to have it added to `libgit2` first.

You can not use `property` as decorator if you don't have a setter since the first param in the `__init__` for `property` is `fget`.

pygit2/settings.py
jdavid: I'll look at the issues this week-end.
But generally speaking, from PEP-8 itself:
`Consistency with this style guide is important. Consistency within a project is more important.`

Though we don't have a project style guide. Well, will check this week-end.

jdavid: The whole paragraph:
`A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is the most important.`

jdavid: By the way, thank you both for the awesome work you're doing!

pygit2/settings.py
webknjaz: But you can track it on the Python side because it's needed to properly set the other value.

pygit2/settings.py
webknjaz: This rule doesn't apply here. Since all other attributes correspond to one value. With two values it doesn't make sense. Setting two values by assigning a tuple to an attribute is ridiculous in Python.

And also when writing Python it doesn't have to look like another language https://www.youtube.com/watch?v=wf-BqAjZb8M. It's fine in the low-level wrapper but not in a public interface.

webknjaz: And I wasn't referring to just PEP8. It's quite a common sense/knowledge. PEP8 is a code style guide but I'm talking about more generic approaches which are probably not formalized explicitly but every Pythonista knows that when they see code looking like Java in `.py` file something isn't quite right.

pygit2/settings.py
webknjaz: you can still use a decorator:
```python
@property
def attr(self):
    """A proper docstring."""
    raise AttributeError

@attr.setter
def attr(self, value):
    # set val
```

pygit2/__init__.py
```diff
@@ -267,3 +270,9 @@ def clone_repository(
     return Repository._from_c(crepo[0], owned=True)
 
 settings = Settings()
+
+try:
+    # try to set default ssl cert locations
+    settings._ssl_cert_locations = default_ssl_verify_paths()
```
webknjaz: I think this belongs at `Settings.__init__()`. It is a good practice to have a ready-to-use object after the initializer run.
Extra initializers are popular in C++, not Python. This semantics is untypical for Pythonic projects.

webknjaz:
I still think that setting one value shouldn't reset another one.

pygit2/settings.py
webknjaz: cc https://github.com/libgit2/pygit2/pull/879#issuecomment-468791003

So here's why we need to track the values we set earlier. Semantically, these setters are assigned with the values for what their names mean.
End-users of the library are Pythonistas who always expect the assignment behavior of the attribute to change exactly the thing its name refers to.
So resetting another setting would be misleading and is a side-effect.
That's why we need to track values which we set earlier for one connected setting to use them when setting the other.
Of course, it's not necessary to expose them for reading right now. But it's a least we can do to preserve consistent UX when setting one of TSL cert location sub-options.
As for setting both sub-options, it should be an explicit method because semantically setting a tuple to attribute to side-effect two values is weird.

imbuedHope: Digging around, it looks like ultimately `libgit2` makes a call to `SSL_CTX_load_verify_locations` when built with Openssl and the [man page](https://linux.die.net/man/3/ssl_ctx_load_verify_locations) for it seems to describe the following behavior.

`When looking up CA certificates, the OpenSSL library will first search the certificates in CAfile, then those in CApath.`

webknjaz: Nice, that's what I thought...

jdavid:
Finally looked at the api. I would replace both setters by a single method:

    set_ssl_cert_locations(...)

webknjaz:
I'd keep properties for individual suboptions. They are native to Python and is what a typical Pythonista would expect. It better communicates the intent as in "make this setting use that value".

jdavid:
I'd rather not have 2 ways to do the same thing. It's good for the users to have 1 way, so they don't have to stop thinking which one they should use, and code looks the same whoever writes it.

It's also good for the developers/maintainers not to increase the burden: our limited resources are the main constraint. Aiming for feature completeness, api consistency, and quality documentation is already ambitious. So I'd keep it simple and stick to the 1 way principle (even if the API feels a bit low level, it's not awkward either).

The `settings.py` file is a good example of the inconsistency we have. It already mixes properties and methods. And a couple of methods even lack the verb at the beginning. `caching` and `cache_max_size` should be setters, and `cache_object_limit` should be renamed to `set_cache_object_limit`

Then the `settings.py` file would be all getter/setters except for the 2 not straightforward cases where we would have methods. The advantages: consistency, simplicity, 1 way.

webknjaz:
Yes, there should be one way of doing things. Except for the cases where that way cannot be used. So are you okay with property-based getters/setters then?

jdavid:
Could you summarize the API you propose for ssl cert locations?

webknjaz:
I'll send it as a PR on top of this branch

webknjaz:
Side note: after looking at `Settings` class I think that this use-case might be a good fit for using `attrs` library. But it's a topic for a separate discussion.

webknjaz:
@jdavid https://github.com/libgit2/pygit2/pull/884

rcoup:
Example if you're using [certifi](https://pypi.org/project/certifi/):

```python
import certifi
import pygit2
pygit2.settings.ssl_cert_file = certifi.where()
```

PR #882, opened 2019-03-01
setup.py: Remove lock on pycparser

yan12125:
Re-submit https://github.com/libgit2/pygit2/pull/867

A local build works. Let's see what will happen on Travis CI.


yan12125:
Mysterious...

PR #883, opened 2019-03-01
Fix testing manylinux1 wheels

webknjaz:


webknjaz:
Here's the build log: https://travis-ci.com/webknjaz/pygit2/jobs/181669779

webknjaz:
@jdavid CI test failures are unrelated and existed in master earlier.

BTW could you please enable CRON in Travis CI?

jdavid:
Do you want to add a cron job to run a build on master periodically? daily, weekly, monthly?
Which is the purpose?

webknjaz:
Purpose is that it could run additional heavy jobs + ensure that dependencies are healthy.

jdavid:
added: master, weekly, do not run if there has been a build in the last 24h

webknjaz:
In this case it won't run jobs which only run on cron sometimes.

jdavid:
ok changed to always run

PR #884, opened 2019-03-02
Expose setting TLS file and dir lookup locations via settings module

### LINKED ISSUE #876: Allow setting GIT_OPT_SET_SSL_CERT_LOCATIONS option in libgit2

webknjaz:
Usage:
```python
import pygit2
pygit2.settings.ssl_cert_file = '/path/to/file'
pygit2.settings.ssl_cert_dir = '/path/to/folder'
del pygit2.settings.ssl_cert_file
pygit2.settings.set_ssl_cert_locations('/path/to/new/file', '/path/to/new/folder')
```

Co-Authored-By: Sriram Raghu <imbuedhope@gmail.com>

Closes #876
Superseeds and closes #879

### DIFF
--- a/pygit2/settings.py
+++ b/pygit2/settings.py
@@ -26,6 +26,8 @@
 # Boston, MA 02110-1301, USA.
 """Settings mapping."""
 
+from ssl import get_default_verify_paths
+
 from _pygit2 import option
 from _pygit2 import GIT_OPT_GET_SEARCH_PATH, GIT_OPT_SET_SEARCH_PATH
 from _pygit2 import GIT_OPT_GET_MWINDOW_SIZE, GIT_OPT_SET_MWINDOW_SIZE
@@ -34,7 +36,7 @@ from _pygit2 import GIT_OPT_SET_CACHE_OBJECT_LIMIT
 from _pygit2 import GIT_OPT_GET_CACHED_MEMORY
 from _pygit2 import GIT_OPT_ENABLE_CACHING
 from _pygit2 import GIT_OPT_SET_CACHE_MAX_SIZE
-
+from _pygit2 import GIT_OPT_SET_SSL_CERT_LOCATIONS
 
 
 __metaclass__ = type  # make all classes new-style by default
@@ -52,9 +54,16 @@ class SearchPathList:
 class Settings:
     """Library-wide settings interface."""
 
-    __slots__ = []
+    __slots__ = '_default_tls_verify_paths', '_ssl_cert_dir', '_ssl_cert_file'
 
     _search_path = SearchPathList()
+
+    def __init__(self):
+        self._default_tls_verify_paths = get_default_verify_paths()
+        self.set_ssl_cert_locations(
+            self._default_tls_verify_paths.cafile,
+            self._default_tls_verify_paths.capath,
+        )
 
     @property
     def search_path(self):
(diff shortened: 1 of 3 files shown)


pygit2/settings.py
```diff
@@ -101,4 +115,43 @@ def cache_object_limit(self, object_type, value):
         """
         return option(GIT_OPT_SET_CACHE_OBJECT_LIMIT, object_type, value)
 
+    @property
+    def ssl_cert_file(self):
+        """TLS certificate file path."""
+        raise AttributeError(
```
webknjaz: @jdavid since we're tracking what we set anyway, we have all values to also expose this for reading. I think it'd be better consistency with this.

imbuedHope:
Might be good to make it clear that `Settings` should not be instantiated by users too. Especially since it's accessible as `pygit2.Settings`. With your changes creating a `Settings` object will cause the ssl certs to reset.

I'm not sure if anyone is using the object, but it's better to play it safe I think. This was not an issue until now since, there was no state was maintained in the `Settings` class.

webknjaz:
Should it be a singleton?

pygit2/settings.py
jdavid: Agree.

jdavid:
Could you please squash the commits to make it easier to review?

imbuedHope:
A singleton could work. I'm not sure if there are scenarios where it makes sense to have multiple `Settings` objects, it doesn't feel like there would be.

webknjaz:
@imbuedHope multiple settings may be useful if you deal with multiple repos with different requirements. But for that, it'd have to be thread-safe. Is it now?

webknjaz:
@jdavid rebased, only atomic commits left.

imbuedHope:
@webknjaz current implementation should be safe with the `multiprocessing` module since memory is **not** shared. I will probably break with `threading`, but I think anyone using it with `threading` should be expecting such behavior.

webknjaz:
@imbuedHope that is true. Though, I believe, libgit2 itself supports and is compiled in thread-safe mode: https://github.com/libgit2/pygit2/blob/f483622/travis/build-manylinux1-wheels.sh#L178.

So to summarize: libgit2 is thread-safe, pygit2 _probably_ isn't. So we don't have to care about that right now.

Can we ignore state-management problem in this PR?

Issue #878, opened 2019-02-24
libgit 0.28 compatible release


0-wiz-0:
Please release a version of pygit2 that's compatible with the current latest release of libgit, 0.28.1.
Thanks.

mwilck:
I'm interested in this one too. The recent update of libgit2 on openSUSE Factory has broken pygit2. I'd appreciate some information when a 0.28.0-compatible release might be available, or what changes are anticipated to make pygit2 work with libgit2 0.28.0.

jayvdb:
The repo here is now using 0.28 since https://github.com/libgit2/pygit2/commit/8dabd78fae7ad7c0de342f2ba5014dceee5ac226, however 11 tests are failing as seen at https://travis-ci.org/libgit2/pygit2/jobs/501118496

These failures correlate with breaking changes in the 0.28 relnotes at https://github.com/libgit2/libgit2/blob/master/docs/changelog.md

jdavid:
There's 1 failure left now:

```
E       GitError: failed to insert entry: invalid object specified - .gitignore
```

If someone wants to give it a try :wink:

webknjaz:
I don't have time to dig into this myself but just to confirm: are you going to create a release once the CI is green?

jdavid:
yes, there's nothing else blocking the release I think

jdavid:
This failure is actually a regression in libgit2, see https://github.com/libgit2/libgit2/issues/5015

@webknjaz @imbuedHope  The tests in AppVeyor are failing as well for a different reason, since PR #884 was merged, see https://ci.appveyor.com/project/jdavid/pygit2/branch/master/job/mwf4pj02fe0yubwk#L7514

```
test\utils.py:42: in <module>
    import pygit2
pygit2\__init__.py:269: in <module>
    settings = Settings()
pygit2\settings.py:65: in __init__
    self._default_tls_verify_paths.capath,
pygit2\settings.py:151: in set_ssl_cert_locations
    option(GIT_OPT_SET_SSL_CERT_LOCATIONS, ssl_cert_file, ssl_cert_dir)
E   GitError: TLS backend doesn't support certificate locations

```

webknjaz:
@jdavid I think, this should probably be turned off under Windows. Another option would be building Windows wheels against OpenSSL.

jdavid:
@fourplusone

webknjaz:
Another solution would be wrapping things with try/except and ignoring just this specific error.

[...]

PR #886, opened 2019-03-09
🚑🐛 Fix TLS locations init when unsupported

webknjaz:
Ignore TLS certificate lookup file and dir locations initialization
failure when the TLS backend library doesn't support it.

Ref #878

### DIFF
--- a/pygit2/settings.py
+++ b/pygit2/settings.py
@@ -38,6 +38,8 @@ from _pygit2 import GIT_OPT_ENABLE_CACHING
 from _pygit2 import GIT_OPT_SET_CACHE_MAX_SIZE
 from _pygit2 import GIT_OPT_SET_SSL_CERT_LOCATIONS
 
+from .errors import GitError
+
 
 __metaclass__ = type  # make all classes new-style by default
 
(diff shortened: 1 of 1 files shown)


jdavid:
Not yet fixed, https://ci.appveyor.com/project/jdavid/pygit2/builds/22951826/job/airgmes6ye5i1vqg#L7623
still failing

But yes, this is the good approach.

webknjaz:
@jdavid can you check this under windows? I don't understand why it doesn't work...

webknjaz:
I missed exception stringification... Updated the commit.

webknjaz:
Alright, it works now.

pygit2/settings.py
```diff
@@ -59,11 +61,23 @@ class Settings:
...
+        try:
+            self.set_ssl_cert_locations(
+                self._default_tls_verify_paths.cafile,
+                self._default_tls_verify_paths.capath,
+            )
+        except GitError as git_err:
+            valid_msg = "TLS backend doesn't support certificate locations"
+            if str(git_err) != valid_msg:
+                raise
+            self._ssl_cert_file = self._default_tls_verify_paths.cafile
```
webknjaz: @jdavid is it okay like this or should we reset everything to `None`?

jdavid: I think it makes more sense to reset everything to `None`

PR #888, opened 2019-03-13
Support for git_merge_analysis_for_ref()

rcoup:
via new `Repository.merge_analysis_for_ref(our_ref, their_head)`

Upstream method was added in [libgit2 v0.28](https://github.com/libgit2/libgit2/releases/tag/v0.28.0)

Code-wise, follows the pattern + tests of the existing `Repository.merge_analysis()`

### DIFF
--- a/src/repository.h
+++ b/src/repository.h
@@ -79,5 +79,7 @@ PyObject* Repository_blame(Repository *self, PyObject *args, PyObject *kwds);
 PyObject* Repository_merge(Repository *self, PyObject *py_oid);
 PyObject* Repository_cherrypick(Repository *self, PyObject *py_oid);
 PyObject* Repository_apply(Repository *self, PyObject *py_diff);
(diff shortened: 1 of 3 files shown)


src/repository.h
```diff
@@ -79,5 +79,7 @@ PyObject* Repository_blame(Repository *self, PyObject *args, PyObject *kwds);
 PyObject* Repository_merge(Repository *self, PyObject *py_oid);
 PyObject* Repository_cherrypick(Repository *self, PyObject *py_oid);
 PyObject* Repository_apply(Repository *self, PyObject *py_diff);
+PyObject* Repository_merge_analysis(Repository *self, PyObject *py_id);
```
rcoup: Couldn't see why `Repository_merge_analysis` was missing here already (oversight?) so I added it.

jdavid:
Merged, thanks.

Probably it would be better to have a single method with an optional parameter, like: `Repository.merge_analysis(id, ref=None)`. Would you like to do a PR to change that?

Issue #814, opened 2018-08-31
pass reference to add_worktree 


Richard-Mathie:
`repo.add_worktree` only has arguments of name and path so it is currently not possible to create a work tree on a specific reference (existing branches, tags ect) and it just creates a new branch `name`

```python
repo.listall_branches()
>>  ['master',
     'develop',
     'foo']

worktree_dir = tempfile.mkdtemp()
os.rmdir(worktree_dir)
worktree_name = 'foo'
worktree = repo.add_worktree(worktree_name, worktree_dir)
>>  Traceback (most recent call last):

      File "<ipython-input-134-cb34d9dc7f6a>", line 4, in <module>
        worktree = repo.add_worktree(worktree_name, worktree_dir)

    ValueError: failed to write reference 'refs/heads/foo': a reference with that name already exists.
```

my current workaround is to move the reference after the fact but this will still create the unwanted branch, and may leave files in the directory.
```python
branch = repo.branches[name]
branch.set_target(repo.branches[the_branch_i_wanted].target)
# or
branch.set_target(repo.references[some reference].target)
```

Richard-Mathie:
further more we get a segfault on doing this on a bare repo ...
```python
>>> import pygit2
>>> repo = pygit2.init_repository('./bare_repo', True)
>>> repo.list_worktrees()
    []
>>> tree =repo.TreeBuilder().write()
>>> commit_id = repo.create_commit('refs/heads/master',
         repo.default_signature, repo.default_signature, "first commit", tree, [])
>>> repo.branches.create('develop', repo[commit_id])
>>> repo.list_worktrees()
    []
>>> repo.listall_branches()
    ['develop', 'master']
```

in bash on the repo
```bash
$ cd bare_repo
$ git branch 
      develop
    * master
$ git worktree add ../versions/develop develop
Preparing ../versions/bare/develop (identifier develop)
    HEAD is now at 4532215 first commit
$ git worktree list
    /..../bare_repo              (bare)
    /..../../versions/develop  4532215 [develop]
```
back in python
```python
>>> repo.list_worktrees()
    ['develop']
>>> repo.lookup_worktree('develop')
    Segmentation fault (core dumped)
```

Any idea why?

jdavid:
[...]

PR #889, opened 2019-03-14
add_worktree: support optional target ref

Insomniac66:
Adds suggestion from #814 by adding an optional positional parameter to `add_worktree`.


jdavid:
Looks good. Could you add a unit test?

PR #893, opened 2019-03-25
Deprecate Patch.patch and add Patch.text and Patch.data

jnrbsn:
This is an attempt to finish up the work in PR #790 originally done by @erikvanzijst. Thanks to him for the initial work.

`Patch.patch` assumes all content to be encoded in UTF-8 and forcefully replaces any non-decodeable sequences. This can lead to corruption for content that either does not conform to any specific encoding altogether, or uses an encoding that is incompatible with, or ambiguous to UTF-8.

As discussed in #790, this change deprecates `Patch.patch` in favor of `Patch.text` and adds `Patch.data`, which returns the unmodified, raw bytes.


jnrbsn:
@jdavid Can you take a look at this? I think I implemented what you described in #790.

pypingou:
I'd argue this is not deprecating, it's purely replacing. Deprecating implies there is a warning and no API breakage (yet).

pypingou:
You could keep ``.patch`` present have it emit a warning and make its output be the one of ``.text``. This would keep backward compatibility and warn users about upcoming changes to the API.

jnrbsn:
@pypingou That's exactly what I already did. `Patch.patch` raises a `DeprecationWarning` and literally just calls `Patch.text`.

pypingou:
Oh, I see it now, sorry about that I had misread the diff initially :(

PR #911, opened 2019-05-23
Download zlib over https.

dbaxa:
Signed-off-by: David Black <dblack@atlassian.com>


dbaxa:
cc @webknjaz (re #875)

PR #935, opened 2019-09-03
Add support for git_remote_ls()

dsully:
Let me know if you need any changes here.

### DIFF
--- a/pygit2/decl/oid.h
+++ b/pygit2/decl/oid.h
@@ -2,6 +2,7 @@ typedef struct git_oid {
 	unsigned char id[20];
 } git_oid;
 
+// This should go to net.h but due to h_order in _run.py, ffi won't compile properly.
 typedef struct git_remote_head {
         int local;
         git_oid oid;
--- a/pygit2/decl/oid.h
+++ b/pygit2/decl/oid.h
@@ -1,3 +1,11 @@
 typedef struct git_oid {
 	unsigned char id[20];
 } git_oid;
+
+typedef struct git_remote_head {
+        int local;
+        git_oid oid;
+        git_oid loid;
+        char *name;
+        char *symref_target;
+} git_remote_head;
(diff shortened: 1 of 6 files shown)


dsully:
Doesn't look like the appveyor issues have anything to do with this change.

pygit2/decl/oid.h
```diff
@@ -1,3 +1,11 @@
...
 	unsigned char id[20];
 } git_oid;
+
+typedef struct git_remote_head {
+        int local;
+        git_oid oid;
+        git_oid loid;
+        char *name;
+        char *symref_target;
+} git_remote_head;
```
jdavid: Please move this to `pygit2/decl/net.h` since in libgit2 this struct is defined in `net.h`, to keep things organized.

test/test_remote.py
```diff
@@ -175,6 +175,14 @@ def test_remote_list(self):
         remote = self.repo.create_remote(name, url)
         assert remote.name in [x.name for x in self.repo.remotes]
 
+    def test_ls_remotes(self):
```
jdavid: Use `@unittest.skipIf(utils.no_network(), "Requires network")`
Tests must pass even when there's no network.

test/test_remote.py
```diff
@@ -175,6 +175,14 @@ def test_remote_list(self):
...
+        assert 1 == len(self.repo.remotes)
+        remote = self.repo.remotes[0]
+
+        refs = remote.ls_remotes()
+
+        assert refs
```
jdavid: May it be possible to test the contents of `refs` a little bit? Like the expected dict keys are present.

pygit2/decl/oid.h
dsully: When I do that, ffi complains: 

`
cffi.CDefError: cannot parse "int git_remote_ls(const git_remote_head ***out, size_t *size, git_remote *remote);"
<cdef source string>:554:25: before: git_remote_head
`

Which I'm guessing is related to this comment/code in pygit2/_run.py:

```
# Order matters
h_files = ...
```

What's the right solution here?

jdavid: ok, no problem then, leave it in `oid.h`, add just a comment like `this should go to net.h but ...`

PR #942, opened 2019-09-16
OdbBackend: initial groundwork

ddevault:
The rough plan is to make the `OdbBackend` type a fairly straightforward mirror of the `odb_backend` in libgit2. I'll then be adding `LooseOdbBackend` and `PackOdbBackend`, and `MempackOdbBackend`, which will inherit from `OdbBackend` and set the underlying `odb_backend` pointer, which should Just Werk based on the approach I have set out so far.

The more interesting case is that of user-created `OdbBackend` implementations. I'm not really sure what to do for this yet. The first inch of this work is done in the form of the read call raising `NotImplementedException` if the C implementation is NULL, then through some kind of magic I assume I'll be able to set these to function pointers for some kind of proxy into Python code with some kind of magic base class. Or something. ¯\\\_(ツ)\_/¯

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -29,10 +29,13 @@
 #include <Python.h>
 #include "error.h"
 #include "object.h"
+#include "odb_backend.h"
 #include "oid.h"
 #include "types.h"
 #include "utils.h"
 #include <git2/odb.h>
+
+PyTypeObject OdbBackendType;
 
 static git_otype
 int_to_loose_object_type(int type_id)
@@ -95,7 +98,9 @@ Odb_build_as_iter(const git_oid *oid, void *accum)
 
     err = PyList_Append((PyObject*)accum, py_oid);
     Py_DECREF(py_oid);
-    return err;
+    if (err < 0)
+        return GIT_EUSER;
+    return 0;
 }
 
 PyObject *
@@ -103,17 +108,20 @@ Odb_as_iter(Odb *self)
 {
     int err;
     PyObject *accum = PyList_New(0);
-    PyObject *ret;
+    PyObject *ret = NULL;
 
     err = git_odb_foreach(self->odb, Odb_build_as_iter, (void*)accum);
     if (err == GIT_EUSER)
-        return NULL;
-    if (err < 0)
-        return Error_set(err);
+        goto exit;
+    if (err < 0) {
+        ret = Error_set(err);
+        goto exit;
+    }
 
     ret = PyObject_GetIter(accum);
-    Py_DECREF(accum);
 
+exit:
+    Py_DECREF(accum);
     return ret;
 }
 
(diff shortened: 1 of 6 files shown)


src/odb_backend.c
```diff
@@ -0,0 +1,142 @@
...
+#include <git2/sys/odb_backend.h>
+
+
+PyDoc_STRVAR(OdbBackend_read__doc__,
+    "read(oid) -> (type, data, size)\n"
+    "\n"
+    "Read raw object data from this odb backend.\n");
+
+PyObject *
+OdbBackend_read(OdbBackend *self, PyObject *py_hex)
```
ddevault: Note, in keeping with the libgit2 API design, I have no intention of refactoring `Odb_read` to go through this code for the sake of DRY.

src/odb_backend.c
```diff
@@ -0,0 +1,142 @@
...
+    #if PY_MAJOR_VERSION == 2
+        "(ns#)",
+    #else
+        "(ny#)",
+    #endif
+        type, data, sz);
+
+    /* XXX: This assumes the default libgit2 allocator is in use and will
+     * probably segfault and/or destroy the universe otherwise */
+    free(data);
```
ddevault: :hankey:

src/odb.c
```diff
@@ -234,6 +237,80 @@ Odb_write(Odb *self, PyObject *args)
...
+    size_t i, nbackends;
+    git_odb_backend *backend;
+    PyObject *accum = PyList_New(0);
+    PyObject *ret;
+
+    nbackends = git_odb_num_backends(self->odb);
+    for (i = 0; i < nbackends; ++i) {
+        err = git_odb_get_backend(&backend, self->odb, i);
+        if (err != 0)
+            return Error_set(err);
```
jdavid: If there's an error you've to `Py_DECREF(accum)` to avoid leaking.

src/odb.c
```diff
@@ -234,6 +237,80 @@ Odb_write(Odb *self, PyObject *args)
...
+PyDoc_STRVAR(Odb_backends__doc__,
+    "Return an iterable of backends for this object database.");
+
+PyObject *
+Odb_backends__get__(Odb *self)
+{
+    int err;
+    size_t i, nbackends;
+    git_odb_backend *backend;
+    PyObject *accum = PyList_New(0);
```
jdavid: Check for errors, if `accum` is `NULL`

jdavid:
looks good
just made a couple of small comments regarding error checking and handling

ddevault:
I think the basics are in good shape here. I'm going to implement the rest of the backend interface next and then figure out how to make new backends in Python in a future pull request.

src/odb_backend.c
```diff
@@ -0,0 +1,348 @@
...
+
+static int
+OdbBackend_build_as_iter(const git_oid *oid, void *accum)
+{
+    int err;
+    PyObject *py_oid = git_oid_to_python(oid);
+
+    err = PyList_Append((PyObject*)accum, py_oid);
+    Py_DECREF(py_oid);
+    return err;
```
jdavid: Return `GIT_EUSER` if there's an error.
`PyList_Append` will return -1 on error, which would be interpreted as `GIT_ERROR` in `OdbBackend_as_iter`, and the exception set by `PyList_Append` would be overwritten.

More information in https://github.com/libgit2/libgit2/issues/4783

`Odb_build_as_iter` has the same problem, please fix it there as well.

src/odb_backend.c
```diff
@@ -0,0 +1,348 @@
...
+    int err;
+    PyObject *accum = PyList_New(0);
+    PyObject *ret;
+
+    err = self->odb_backend->foreach(self->odb_backend,
+            OdbBackend_build_as_iter, (void*)accum);
+    if (err == GIT_EUSER)
+        return NULL;
+    if (err < 0)
+        return Error_set(err);
```
jdavid: On these 2 errors you've to `Py_DECREF(accum);`.
Same problem in `Odb_as_iter`

src/odb_backend.c
```diff
@@ -0,0 +1,348 @@
...
+    err = PyList_Append((PyObject*)accum, py_oid);
+    Py_DECREF(py_oid);
+    return err;
+}
+
+PyObject *
+OdbBackend_as_iter(OdbBackend *self)
+{
+    int err;
+    PyObject *accum = PyList_New(0);
```
jdavid: Check for errors returned by `PyList_New`
Same problem in Odb_as_iter

src/odb_backend.c
```diff
@@ -0,0 +1,348 @@
...
+        Error_set_oid(err, &oid, len);
+        return NULL;
+    }
+
+    tuple = Py_BuildValue(
+    #if PY_MAJOR_VERSION == 2
+        "(ns#)",
+    #else
+        "(ny#)",
+    #endif
```
jdavid: I've removed support for Python 2 in master, so you can update this.

src/odb_backend.c
```diff
@@ -0,0 +1,348 @@
...
+                        "OdbBackendLoose takes no keyword arguments");
+        return -1;
+    }
+
+    const char *fmt =
+    #if PY_MAJOR_VERSION == 2
+        "Oii|II";
+    #else
+        "Oip|II";
+    #endif
```
jdavid: I've removed support for Python 2 in master, so you can update this.

jdavid:
I've fixed the build errors in AppVeyor, now there're some test failures related to this work.

ddevault:
I've addressed the feedback and I think that this is mergable, except for the build failure. I looked into it and I don't really understand why it's occuring. By my reckoning, this should affect a lot of tests. Can you take a look and tell me what you think?

jdavid:
We've seen these errors before.

Looking for a little bit I've found commit 00dd78bf1b3 which fixed 7 errors (6 of them were PermissionErrors): https://ci.appveyor.com/project/techtonik/pygit2/build/1.0.12/tests

Commit af983b9 is likely related as well.

ddevault:
Thanks!

PR #959, opened 2019-11-28
Fix build, change git_attr_t to git_attr_value_t

lberk:
When using libgit2 built from source, the build breaks due to a change
from git_attr_t to git_attr_value_t.  Update our headerfile
accordingly.


jdavid:
Thanks, then we've to wait for the next release of libgit2, I believe it will be 0.99, hopefully it will not take long.

Issue #877, opened 2019-02-21
Add binding for git_diff_patchid


sffc:
This function was added to libgit2 recently (v0.27.0), but I can't find a Python binding for it in pygit2.  I would like to be able to use it.

https://libgit2.org/libgit2/#HEAD/group/diff/git_diff_patchid

We would like this feature to use in ICU tooling: [ICU-20444](https://unicode-org.atlassian.net/browse/ICU-20444)

sffc:
We would like this feature to use in ICU tooling: [ICU-20444](https://unicode-org.atlassian.net/browse/ICU-20444)

jdavid:
Done in PR #960

PR #960, opened 2019-11-28
diff: add support for patchid.

herrerog:
Add binding for git patchid as mentioned in issue #877 
For the test to pass, it must be built against `fb439c975` in libgit2 where support for patchid has been improved.


jdavid:
Thanks, then we've to wait for the next release of libgit2, I believe it will be 0.99, hopefully it will not take long.

PR #982, opened 2020-02-28
Introduce Refdb type

ddevault:
TODO:

- [x] Wrap git_refdb in Refdb type
- [x] Add Reference constructor(s)
- [x] Create RefdbBackend
- [x] Create repositories with custom refdb/odb
- [x] Create Refdb from scratch
- [x] Tests
- [x] Update `docs/backends.rst`

Future work:

- Iterating over a refdb_backend from Python
- First-class reflog support
- Transactional refdb updates

### DIFF
--- a/src/odb_backend.c
+++ b/src/odb_backend.c
@@ -772,8 +772,8 @@ PyTypeObject OdbBackendPackType = {
 };
 
 PyDoc_STRVAR(OdbBackendLoose__doc__,
-        "OdbBackendLoose(objects_dir, compression_level,\n"
-        "    do_fsync, dir_mode=0, file_mode=0)\n"
+        "OdbBackendLoose(objects_dir, compression_level,"
+        " do_fsync, dir_mode=0, file_mode=0)\n"
         "\n"
         "Object database backend for loose objects.\n"
         "\n"
--- a/src/odb_backend.c
+++ b/src/odb_backend.c
@@ -65,31 +65,34 @@ static int
 pygit2_odb_backend_read(void **ptr, size_t *sz,
         git_object_t *type, git_odb_backend *_be, const git_oid *oid)
 {
-    int err;
     PyObject *args, *py_oid, *result;
     struct pygit2_odb_backend *be = (struct pygit2_odb_backend *)_be;
 
     py_oid = git_oid_to_python(oid);
-    args = Py_BuildValue("(O)", py_oid);
-    result = PyObject_CallObject(be->read, args);
-    Py_DECREF(py_oid);
+    if (py_oid == NULL)
+        return GIT_EUSER;
 
-    if ((err = git_error_for_exc()) != 0) {
-        return err;
-    }
+    args = Py_BuildValue("(N)", py_oid);
+    result = PyObject_CallObject(be->read, args);
+    Py_DECREF(args);
 
     if (result == NULL)
-        return GIT_EUSER;
+        return git_error_for_exc();
 
     const char *bytes;
-    if (!PyArg_ParseTuple(result, "ny#", type, &bytes, sz) || !bytes)
+    if (!PyArg_ParseTuple(result, "ny#", type, &bytes, sz) || !bytes) {
+        Py_DECREF(result);
         return GIT_EUSER;
+    }
 
     *ptr = git_odb_backend_data_alloc(_be, *sz);
-    if (!*ptr)
+    if (!*ptr) {
+        Py_DECREF(result);
         return GIT_EUSER;
+    }
 
     memcpy(*ptr, bytes, *sz);
+    Py_DECREF(result);
     return 0;
 }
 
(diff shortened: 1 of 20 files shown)


src/refdb_backend.c
```diff
@@ -0,0 +1,419 @@
...
+    self->refdb_backend = (git_refdb_backend *)be;
+    return 0;
+}
+
+void
+RefdbBackend_dealloc(RefdbBackend *self)
+{
+    if (self->refdb_backend && self->refdb_backend->free == pygit2_refdb_backend_free) {
+        struct pygit2_refdb_backend *be = (struct pygit2_refdb_backend *)self->refdb_backend;
+        free(be);
```
jdavid: Here we need to clear as well the references to the methods, like it's done in the object db backend code: `Py_CLEAR(be->exists)` etc.

src/refdb_backend.c
```diff
@@ -0,0 +1,419 @@
...
+    Repository *repo = NULL;
+    if (!PyArg_ParseTuple(args, "O", &repo))
+        return -1;
+
+    if (!PyObject_IsInstance((PyObject *)repo, (PyObject *)&RepositoryType)) {
+        PyErr_SetString(PyExc_TypeError,
+                        "Refdb expects an object of type "
+                        "pygit2.Repository");
+        return 0;
+    }
```
jdavid: Same as in the previous commit, use `O!` to make this shorter.

src/refdb_backend.c
```diff
@@ -0,0 +1,419 @@
...
+    if (PyObject_HasAttrString((PyObject *)self, "exists")) {
+        be->exists = PyObject_GetAttrString((PyObject *)self, "exists");
+        be->backend.exists = pygit2_refdb_backend_exists;
+        Py_INCREF(be->exists);
+    }
+
+    if (PyObject_HasAttrString((PyObject *)self, "lookup")) {
+        be->lookup = PyObject_GetAttrString((PyObject *)self, "lookup");
+        be->backend.lookup = pygit2_refdb_backend_lookup;
+        Py_INCREF(be->lookup);
```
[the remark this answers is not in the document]
ddevault: -1 to a macro. I will fix the leaks, though. I think the same problem exists in odb backend.

jdavid:
Looking at the API level.

I see there're a number of ways to get a backend:

```python
refdb = repo.refdb       # wraps git_repository_refdb
refdb = Refdb(repo)      # wraps git_refdb_new
refdb = Refdb.open(repo) # wraps git_refdb_open
```

For consistency and to be explicit please change `Refdb(repo)` to `Refdb.new(repo)`

src/refdb_backend.c
```diff
@@ -102,6 +105,36 @@ pygit2_refdb_backend_lookup(git_reference **out,
...
+    who = build_signature(NULL, _who, "utf-8");
+    old = git_oid_to_python(_old);
+
+    args = Py_BuildValue("(OOOsOs)", ref,
+            force ? Py_True : Py_False,
+            who, message, old, old_target);
+    PyObject_CallObject(be->write, args);
+    Py_DECREF(ref);
+    Py_DECREF(who);
+    Py_DECREF(old);
```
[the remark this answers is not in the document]
ddevault: If `Py_BuildValue` increments the refcount on `O`, and we use `N` instead, do we have to Py_DECREF our references to the ref, signature, and oid? Or will decref'ing the built value (args) cause those to be freed as well?

ddevault:
> For consistency and to be explicit please change `Refdb(repo)` to `Refdb.new(repo)`

Done.

ddevault:
Related libgit2 bug:

https://github.com/libgit2/libgit2/issues/5449

This prevents us from being able to create a RefdbFsBackend and use it to create a Refdb from scratch. Other than that, the functionality of this patch is complete. I'll add tests in a day or two.

src/odb_backend.c
```diff
@@ -185,18 +204,21 @@ pygit2_odb_backend_exists(git_odb_backend *_be, const git_oid *oid)
...
     result = PyObject_CallObject(be->exists, args);
     Py_DECREF(py_oid);
 
     if ((err = git_error_for_exc()) != 0) {
+        Py_DECREF(result);
         return err;
     }
 
     if (!result)
         return GIT_EUSER;
```
jdavid: This code reads different than the code before, but the errors are more or less the same:

- Not checking `git_oid_to_python` and `Py_BuildValue` for errors
- Call to `Py_DECREF(NULL)`
- `if (!result)` always false

Also here you're not decref'ing `args`.

ddevault:
Addressed comments on odb fixes. Still need to write tests and verify that everything works. I ran into some issues in a shallow test the other day.

ddevault:
Note: Windows CI is failing due to lack of [fnmatch](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fnmatch.html) support, which I used to implement globbing.

Options:
- Drop in musl libc's implementation, ~300 LoC, MIT licensed
- Add a code path for Windows which doesn't support globbing
- Implement globbing some other way

ddevault:
✓ Tests

Code-wise, this should be ready for review. The question of what to do about fnmatch on Windows remains, and I need to write up some docs.

ddevault:
Oh, note that this requires the following libgit2 change before it'll work: https://github.com/libgit2/libgit2/pull/5456

jdavid:
Regarding `fnmatch` does libgit2 provide an implementation for this? Apparently they did, but replaced it by wildmatch, see https://github.com/libgit2/libgit2/pull/5110 So I think we should use the same as libgit2.

ddevault:
https://github.com/libgit2/libgit2/issues/5458

I'll just vendor it in the meanwhile (on Monday).

ddevault:
✓ fnmatch -> wildmatch

✓ Questions answered from libgit2 about the semantics of refdb_backend.write

Still needs docs but otherwise I think this is good to go.

jdavid:
The test failures are because of https://github.com/libgit2/libgit2/pull/5456 ?
I see this has been merged to `master` but not to `maint/v0.99`, then please update the CI files to use master, so the tests pass.

ddevault:
Got all of the CI pointing at libgit2 master and the tests are passing.

ddevault:
I have rigged up the documentation as well.

jdavid:
Cool, now we have to wait for a libgit2 release with https://github.com/libgit2/libgit2/pull/5456
Hopefully it won't take long.

ddevault:
I pushed a little bit of clean up in the course of implementing this:

https://git.sr.ht/~sircmpwn/git.sr.ht/commit/a3fe57b6844709eac64fdca60370e5c227979fd3

It works :grin:

src/refdb_backend.c
```diff
@@ -198,7 +198,6 @@ pygit2_refdb_backend_lookup(git_reference **out,
 
     *out = result->reference;
 out:
-    Py_DECREF(result);
```
jdavid: Why this? Doesn't `lookup(..)` return a new reference?

ddevault: Apparently not. This was causing double frees.

ddevault:
Thanks!

jdavid:
Thank you!

PR #988, opened 2020-03-13
macOS Wheels

rcoup:
Add macOS wheels into CI, using Github Actions macOS support.

* Built against Python 3.6, 3.7, 3.8 using the Python.org distributions, which means wheels are compatible with macOS 10.9+.
* Wheels run through [delocate](https://github.com/matthew-brett/delocate/) to embed libgit2
* Unit tests run/pass against the wheels
* Caches things fairly heavily, so total rebuild time is normally 2-3mins.
* Wheels are saved as artifacts, and are downloadable for 90 days.

### Libgit2

Builds against `maint/v0.99` to match appveyor & travis, but it's configured as a matrix so can easily test against releases/branches/etc if appropriate. Not sure if we should be doing a "release" build `-DCMAKE_BUILD_TYPE=Release` (or `RelWithDebugInfo`), but it matches the other environments atm.

### Python 3.5

This was in the plan, but I think the Python.org 3.5.4 release was built using an old version of GCC instead of Clang, and it's not in the CI environment by default. I can dig further, or switch to the build of Py3.5 used by GH Actions (which is macOS 10.13+ only), or we just leave it out.

### PyPi

Haven't set it up to push to PyPi on release tags, but that's easily done using eg: [gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) once secrets are configured.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,119 @@
+---
+name: macOS
+
+on:
+  push: {}
+
+jobs:
+  build:
+    name: py-${{ matrix.py.ver }} libgit2-${{ matrix.libgit2}}
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        libgit2: ['maint/v0.99']
+        py:
+          - ver: '3.6'
+            release: '3.6.8'  # last Python.org binary release
+          - ver: '3.7'
+            release: '3.7.7'
+          - ver: '3.8'
+            release: '3.8.2'
+    env:
+      HOMEBREW_CACHE: ${{ github.workspace }}/.cache/brew
+      CCACHE_DIR: ${{ github.workspace }}/.cache/ccache
+      CCACHE_COMPRESS: 1
+      PIP_CACHE_DIR: ${{ github.workspace }}/.cache/pip
+    steps:
+      - name: checkout pygit2
+        uses: actions/checkout@v2
+        with:
+          path: pygit2
+
+      - name: checkout libgit2
+        uses: actions/checkout@v2
+        with:
+          repository: libgit2/libgit2
+          path: libgit2
+          ref: ${{ matrix.libgit2 }}
+
+      - id: cache
+        uses: actions/cache@v1.1.0
+        with:
+          path: .cache
+          key: cache-libgit2.${{ matrix.libgit2 }}-py.${{ matrix.py.release }}
+          restore-keys: |
+            cache-
+
+      - name: setup python
+        run: |
+          mkdir -p .cache/python/ .cache/pip/
+          PKG="python-${{ matrix.py.release }}-macosx10.9.pkg"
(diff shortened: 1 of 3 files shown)


jdavid:
I think we can leave out Python 3.5, probably we will drop support for it this year.

Thanks!

.github/workflows/macos.yml
```diff
@@ -0,0 +1,119 @@
...
+          echo "::set-output name=version::$VERSION"
+
+      - name: build pygit2
+        id: pygit2
+        working-directory: pygit2
+        env:
+          LIBGIT2: "${{ github.workspace }}/libgit2/env"
+          LDFLAGS: "-Wl,-rpath,'${{ github.workspace }}/libgit2/env/lib'"
+        run: |
+          python3 setup.py bdist_wheel
```
webknjaz, line 92: FTR the recommended way of doing this is something like `pip wheel --no-deps -w dist .`

.github/workflows/macos.yml
```diff
@@ -0,0 +1,119 @@
...
+  build:
+    name: py-${{ matrix.py.ver }} libgit2-${{ matrix.libgit2}}
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        libgit2: ['maint/v0.99']
+        py:
+          - ver: '3.6'
+            release: '3.6.8'  # last Python.org binary release
```
webknjaz, line 17: FTR here's a way to select the concrete "patch" version intelligently: https://github.com/amol-/dukpy/blob/54699a3/.travis.yml#L56-L109

rcoup: Thanks - I need to select the last version available as a binary release from Python.org, not the latest patch version though. I don’t understand why they’re not built automatically, but they’re not.

.github/workflows/macos.yml
rcoup: Same outcome though afaik?

.github/workflows/macos.yml
rcoup: Ah, re-reading, it does that, yes! Not sure it’s a big deal for us atm - we’re not typically using cutting edge CPython features

webknjaz: Yep. Sometimes they release 3.6.8 for Linux, but stop at 3.6.6 for macOS. so this is something that my script accounts for. But yes, it shouldn't matter much. OTOH when you do exit-checks after the build but before publishing the artifacts, it may make sense to run tests on the recent version.

*/me needs to find time and contribute that to cimultibuild...*

.github/workflows/macos.yml
webknjaz: Historic notes:
- https://packaging.python.org/en/latest/guides/modernize-setup-py-project/
- https://packaging.python.org/en/latest/discussions/setup-py-deprecated/
- https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

Issue #895, opened 2019-03-31
Consistent interface to get text and bytes


jdavid:
Follow up from #610 #790 and #893

General policy:

- Do not implement `str()`
- To get unicode string use `.text`
- In `.text` use UTF-8 and *replace* (the rationale for *replace* is explained in https://github.com/libgit2/pygit2/pull/790#issuecomment-385906316)
- To get the byte string use `.data` or `.raw` (*this is to be decided*)
- For attributes the name of the attribute returns text, prefix with `raw_` to get bytes. For instance `Signature.name` and `Signature.raw_name`
- Implement the buffer protocol, `bytes(..)` where appropriate

Open for discussion.

TODO:

- [ ] Replace `TreeEntry._name` by `.raw_name`
- [ ] Replace `DiffLine.content` by `.text`
- [ ] Inventory all the places where we get bytes, text, or the buffer protocol
- [ ] Settle on `.data` or `.raw`
- [ ] Replace `DiffLine.raw_content` by `.data` or `.raw`
- [ ] Settle on `str()` `bytes()` and the buffer protocol

The case of Oid, what we've now:

- `oid.raw` returns the byte string (that's good, unless we decide to settle on `.data`)
- `str(oid)` and `oid.hex` both return the hex representation, always `<str>` (bytes in Python 2 and text in Python 3)
- Oid is the only place where we implement `str(...)`
- `Object.hex` and `TreeEntry.hex` behave the same, they return always `<str>`. Apparently these are the only places where we always return `<str>`.

jdavid:
If I'm not missing anything..

These are the places where we use `.data`:

- `Blob.data`
- `Patch.data`

These are the places where we use `.raw`:

- `Oid.raw`

With `read_raw()`:

- `Object.read_raw()`

jnrbsn:
Just curious, are you wanting to get all this done _before_ doing another release? How do you decide when to do a release?

jnrbsn:
I prefer `data` as an attribute name, because it's a noun, and it straightforwardly tells you what you're getting. `raw` is an adjective. Raw what? I think having attributes like `raw_name` makes sense, but the word "raw" by itself makes less sense. Also, since the word "data" is already kind of generic, it already kind of implies that no assumptions are being made about the structure of the data, and so it already implies "raw".

jdavid:
Good point.

No, this doesn't block the release. What I'd like to do before the next release is handling the pycparser issue, #846 

[...]

PR #990, opened 2020-03-18
Accept Pathlike objects (eg. pathlib.Path) in Py3.6+

rcoup:
PEP-519 (Py3.6+ / PyPy 7.3+) adds a [file system path protocol[(https://www.python.org/dev/peps/pep-0519/) via `Object.__fspath__()`, and is supported by `pathlib`/`os`/`open()`/etc. This PR adds support for accepting `pathlib.Path` objects natively in Pygit2.

* Where we take paths in Python, check `hasattr(value, '__fspath__')` to provide Py3.5 compatibility, then call it to get a str/bytes representation.
* In C, use [`PyUnicode_FSConverter`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FSConverter) when parsing function arguments (it handles Pathlike objects), or [`PyOS_FSPath()`](https://docs.python.org/3/c-api/sys.html#c.PyOS_FSPath) when available (ie. not Py3.5).
* Implementation-wise, this approach is basically called for all string arguments, but since `os.fspath()`/`PyOS_FSPath()` return a str/bytes argument unchanged, behaviour-wise it's the same. Cleaning this up might sit with #895?

Adds a pile of `*_aspath()` tests under Py3.6+ to wherever we seem to accept paths. I'm sure I've missed some but we can sort that along the way.

### DIFF
--- a/pygit2/index.py
+++ b/pygit2/index.py
@@ -78,7 +78,7 @@ class Index(object):
 
     def __getitem__(self, key):
         centry = ffi.NULL
-        if isinstance(key, str):
+        if isinstance(key, str) or hasattr(key, '__fspath__'):
             centry = C.git_index_get_bypath(self._index, to_bytes(key), 0)
         elif isinstance(key, int):
             if key >= 0:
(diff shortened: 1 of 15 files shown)


jdavid:
AppVeyor is failing with a C syntax error. Travis is failing as well, but now it's under maintenance and cannot see the details.

I think this is the right time to drop support for Python 3.5 and simplify the PR a bit.

jdavid:
So I dropped 3.5 support. Could you please rebase and update the PR? Thanks.

rcoup:
yup, will do

rcoup:
Everything was still needed for CI's version of PyPy, so I left it in for the moment to check tests.

jdavid:
Ok.. I see this would require pypy 7.3.0 while in Travis the latest version is 7.1.1

Issue #830, opened 2018-10-20
Review exceptions


jdavid:
To have one exception for every `GIT_EXXX` all of them inheriting from `GitError`, use a consistent naming, e.g. `GIT_EINVALIDSPEC` would become `GitInvalidSpecError` or maybe just `InvalidSpecError`.

To provide backwards compatibility use multiple inheritance, e.g. `InvalidSpecError` would inherit from `GitError` and `ValueError`. Eventually the base class for backwards compatibility would be remove in some future version.

To settle on naming post here the full mapping of `GIT_EXXX` to exception, even if this can be implemented in steps.

This follows up on issues #531 and #828

robinst:
Sounds good to me.

> Eventually the base class for backwards compatibility would be remove in some future version.

There's some precedent in Python's standard library for exceptions with multiple base classes, see this post:

https://www.reddit.com/r/Python/comments/1k9ygb/multiple_inheritance_for_library_exceptions/

So maybe it's not necessarily a bad thing that needs to be removed in the future.

jdavid:
yes you're right, it may be a good thing

m451:
While I implemented a first version there are currently some reasons that stop us from fully switching to such a system that will probably need to be tackled one-by-one.

- [ ] `Passthrough` needs to be changed to `GitPassthroughError` to be consistent in the naming. The issue is that while I replaced all instances in PyGit2 with the new name, `Passthrough` might be used by users. I therefor let `GitPassthroughError` inherit from `Passthrough` and didn't delete `Passthrough` yet. So, `Passthrough` can be viewed as deprecated and some time in future should be removed.
- [ ] `GIT_EUSER` is currently used to control the flow of some wrappers. It seems the current logic that uses `GIT_EUSER` return codes needs to be changed in a way that it's using the new `GitUserError` exception. This seems to be quite hard as I yet don't fully understand the flow logic there. Some help with this would be awesome.
[...]

Issue #996, opened 2020-04-03
RemoteCallbacks exceptions are ignored


m451:
Currently if one implements a custom `RemoteCallbacks.get_credentials()` and in the process of returning the credentials to the caller throws any Exception, that Exception is ignored and will be replaced by a generic `GitError` exception, limiting any custom exception handling.

The code that does this seems to be in the [remote.py](https://github.com/libgit2/pygit2/blob/master/pygit2/remote.py#L309). To be more precise [here](https://github.com/libgit2/pygit2/blob/be32579ebbe198c637215140454a9bceeeca0d4f/pygit2/remote.py#L309) in the RemoteCallbacks class.

```Python
try:
	ccred = get_credentials(credentials, url, username, allowed)
	cred_out[0] = ccred[0]
except Passthrough:
	return C.GIT_PASSTHROUGH
except Exception as e:
	self._stored_exception = e
	return C.GIT_EUSER
```

It stores the error into `_stored_exception` but simply returns `-7` to the caller. 
The calling function then usually will pass that error code to the generic `check_error()` function of the `errors.py` module. However, the actual Exception stored in `_stored_exception` is never processed there. Instead, it simply raises a `GitError` with the message `err -7 (no message provided)`.

It's unclear how a custom RemoteCallbacks is supposed to perform exception handling when the desired handling is to "bubble the error upwards the stack" where it would be handled and it's unclear why the exception is stored but never re-raised or atleast it's message used to inform the caller about the actual error.

E.g. if someone implements RemoteCallbacks like this:

```Python
class MyCallbacks(RemoteCallbacks):
    def credentials(self, url, username_from_url, allowed_types):
        raise Exception("handle me higher up in the stack!")
```

and calls it like this:

```Python
try:
    pygit2.clone_repository("ssh://github.com/libgit2/pygit2", "pygit2.git", callbacks=MyCallbacks())
except GitError ex:
    print(ex)
except Exception as ex:
    print(f"custom exception handled here: {ex}")
```

One would expect the output to be:
> custom exception handled here: handle me higher up in the stack!

While we actually get:

> err -7 (no message provided)

m451:
As far as I can tell it seems like there would be a bigger refactoring needed to fix this issue propperly. 

So what about a temporary workaround: implement a `CustomPygitError` Exception that users can throw if they wish to handle those errors on their own and, if `CustomPygitError` is detected in the `check_error()` handler just re-raise it.

jdavid:
[...]

PR #998, opened 2020-04-05
Replaced exception handling logic

omniproc:
As discussed in #830 and #996 this is a initial implementation of a new exception handling mechanism. The goals so far have been: 

- Don't break backwards compatibility
- Wrap C-Calls that return an exit code more elegant in a way that a propper Python exception is rised if the exit code is not OK
- If no propper exit code was found raise a generic `GitException` and try to populate it with as much information as available.
- If a native Python exception occurres during that C-call, raise that exception and do not hide it

As suggested / discussed in #830 it was implemented using mutliple inheritance. 
Currently the tree looks like this `Exception -> GitError (for backward compatibility) -> BaseGitException (for multi inheritance support and new exception handling logic) -> GitException (as actual class to derive from) -> GitXxxError (Exceptions matching GIT_E* exit codes from libgit2)`.

The new code is in [errors.py](https://github.com/m451/pygit2/blob/6d45967f4939ee97533c37dbb3c9e744a1cdf40a/pygit2/errors.py) - the other classes have only been touched to implement it. A single unittest was changed to be more specific when detecting the returned exception, thought that change should not break anything.

There are multiple ways to use it.

```Python
@GitException.check_result
def foo(val):
    if val == 99:
        raise Exception("Native Python exception thrown inside foo().")
    return val 

foo(1)      # OK
foo(0)      # OK
foo(-3)     # Raises GitNotFoundError because GIT_ENOTFOUND == -3
foo(-16)    # Raises GitException because GIT_EAUTH == -16 and GitAuthError has not yet been explicitly implemented as standalone Exception
foo(99)     # Raises Exception with text "Native Python exception thrown inside foo().". Don't wrap/hide/manipulate Python exception but just re-raise them!

# If we want to raise a more specific exception in case an error occures we can do so:
@GitNotFoundError.check_result
def bar(val):
    if val == 99:
        raise Exception("Native Python exception thrown inside foo().")
    return val 

bar(1)      # OK
bar(0)      # OK
bar(-3)     # Raises GitNotFoundError because we explicitly told to raise GitNotFoundError for all error codes
bar(-16)    # Raises GitNotFoundError because we explicitly told to raise GitNotFoundError for all error codes
bar(99)     # Raises Exception with text "Native Python exception thrown inside foo().". Don't wrap/hide/manipulate actual Python exception but just re-raise them!

# Also we have control over the message that should be returned to the user
@GitException.check_result(message="Something bad happend at baz().")
def baz(val):
    if val == 99:
        raise Exception("Native Python exception thrown inside foo().")
    return val 

baz(1)      # OK
baz(0)      # OK
baz(-3)     # Raises GitNotFoundError because GIT_ENOTFOUND == -3. Exception message will be "Something bad happend at baz()."
baz(-16)    # Raises GitException because GIT_EAUTH == -16 and GitAuthError has not yet been explicitly implemented as standalone Exception. Exception message will be "Something bad happend at baz()."
baz(99)     # Raises Exception with text "Native Python exception thrown inside foo().". Don't wrap/hide/manipulate actual Python exception but just re-raise them!


# If we don't want to wrap the hole function - thought a redesign in such a way that C-calls are wrapped into individual Python functions would make sense - we can use old style wrapping without decorators.
# This is usefull for a quick replacement of the current exception logic. For example:
GitException.check_result(baz, message="Something bad happend at baz().")(1)
```

### DIFF
--- a/pygit2/__init__.py
+++ b/pygit2/__init__.py
@@ -33,7 +33,7 @@ from ._pygit2 import *
 from .blame import Blame, BlameHunk
 from .config import Config
 from .credentials import *
-from .errors import check_error, Passthrough
+from .errors import GitException, GitPassthroughError, Passthrough
 from .ffi import ffi, C
 from .index import Index, IndexEntry
 from .remote import Remote, RemoteCallbacks, get_credentials
@@ -145,8 +145,7 @@ def init_repository(path, bare=False,
 
     # Call
     crepository = ffi.new('git_repository **')
-    err = C.git_repository_init_ext(crepository, to_bytes(path), options)
-    check_error(err)
+    GitException.check_result(C.git_repository_init_ext)(crepository, to_bytes(path), options)
 
     # Ok
     return Repository(to_str(path))
@@ -250,17 +249,11 @@ def clone_repository(
         callbacks = RemoteCallbacks()
 
     callbacks._fill_fetch_options(opts.fetch_opts)
-
-    err = C.git_clone(crepo, to_bytes(url), to_bytes(path), opts)
-
+    GitException.check_result(C.git_clone)(crepo, to_bytes(url), to_bytes(path), opts)
     # Error handling
     exc = d.get('exception', callbacks._stored_exception)
     if exc:
         raise exc
-
-    check_error(err)
-
-    # Ok
     return Repository._from_c(crepo[0], owned=True)
 
 
--- a/pygit2/config.py
+++ b/pygit2/config.py
@@ -26,7 +26,7 @@
 from cached_property import cached_property
 
 # Import from pygit2
-from .errors import check_error
+from .errors import GitException, GitIOError
 from .ffi import ffi, C
 from .utils import to_bytes
 
@@ -50,9 +50,7 @@ class ConfigIterator(object):
 
     def _next_entry(self):
         centry = ffi.new('git_config_entry **')
-        err = C.git_config_next(centry, self._iter)
-        check_error(err)
-
+        GitException.check_result(C.git_config_next)(centry, self._iter)
         return ConfigEntry._from_c(centry[0], self)
 
     def next(self):
@@ -75,12 +73,10 @@ class Config(object):
         cconfig = ffi.new('git_config **')
 
         if not path:
-            err = C.git_config_new(cconfig)
+            GitIOError.check_result(C.git_config_new)(cconfig)
         else:
             assert_string(path, "path")
-            err = C.git_config_open_ondisk(cconfig, to_bytes(path))
-
-        check_error(err, io=True)
+            GitIOError.check_result(C.git_config_open_ondisk)(cconfig, to_bytes(path))
         self._config = cconfig[0]
 
     @classmethod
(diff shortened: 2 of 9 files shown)


jdavid:
Let's see an exemple. Today we have:
  
```python
err = C.git_clone(crepo, to_bytes(url), to_bytes(path), opts)
check_error(err)
```
  
The proposal is to write:

```python  
GitException.check_result(C.git_clone)(crepo, to_bytes(url), to_bytes(path), opts)
```
  
Today's code already transforms the error code to a Python exception, and it's in my opinion easier to read.
  
The thing the new code does is to catch and re-raise an exception from `C.git_clone`. But `C.git_clone` cannot raise any exception, only return an error code. If `C.git_clone` could raise an exception, it would work already with today's code.

What happens, as seen in issue #996, is that a callback written in Python may raise an exception. Then the callback has to store that exception somewhere (`_stored_exception`), and return `GIT_EUSER`.

For the record, from libgit2 sources:

```
        /**
         * GIT_EUSER is a special error that is never generated by libgit2
         * code.  You can return it from a callback (e.g to stop an iteration)
         * to know that it was generated by the callback and not by libgit2.
         */
        GIT_EUSER      = -7,
```

Then the code that has called `C.git_clone` must check `_stored_exception` and raise an exception. It would be nice to have a more general solution for this, I understand this is one of the stated goals of this PR:

> - If a native Python exception occurres during  that C-call, raise that exception and do not hide it

But the PR doesn't address this actually. The code handling `_stored_exception` is still there.

It looks to me that this PR is actually trying to resolve issue #830, not #996. Am I right?

m451:
> It looks to me that this PR is actually trying to resolve issue #830, not #996. Am I right?

Yes, absolutely. Maybe I should have been more clear about that. When looking into the problem as discussed in #996 I realized that before we can handle that it would make sense to first have a more generic exception handling implemented as suggested in #830. The thing that this PR add's, I hope, is to implement it and thus provides us a *single place* where all errors - Python and C type - can be handled.

Because of this we also get a single place where special error codes like `GIT_EUSER` can be handled differently. For now it'll just convert it into a Python exception. But this could easily be modified in such a way that it does store and - then later, when the generic GitError occures, rises the actual Exception it stored before. Essentially the logic there can be the same as before: get any exception that occures. If it's a C error, store it and return it's value. The code higher up in the stack, from my understanding, will then raise a generic GitError. At that point the implementation could be changed so it don't simply rise the GitError but rise the stored Exception.
The difference is that now that logic, when needed, can be very different for different error types and is still kept in a central place where you'd expect it: in the error module that's supposed to handle all the errors the way we want it.

I have not yet integrated any logic that would specially handle `GIT_EUSER` but that would be the next step to fix #998. I just wanted to better understand what's happening there before I apply the changes, so I opend the todos in #830.
If needed I can supply a working example how the handling of such special errors could work with this PR but I'll need some time to fully understand the call stack and think about the different ways that could be handled.

jdavid:
Okay. The callback code was contributed long time ago by someone else, I myself didn't know exactly how it worked. So I've reviewed it and done some commits. It's all internal refactoring, no external changes (not intended at least).

Now it should be much easier to read and maintain, at least now I understand how it works. You can give a look at the new `pygit2/callbacks.py` file, which now holds all code related to callbacks, it has some documentation strings and comments that should make it easier.

I don't think there's any dependency between issues #830 and #996. There're many calls to `check_error` but only a handful of callbacks, `GIT_EUSER` and `GIT_PASSTHROUGH` are only of interest in the context of callbacks.

Next I'll work on the unit tests concerning #996 (not completely sure, but I think that issue is already fixed).

PR #1009, opened 2020-05-20
Add repr, str, == and != support to IndexEntry.

olsen232:
IndexEntry is a simple datatype which can be usefully displayed or
compated to any other IndexEntry.
Not done for larger types Index or ConflictCollection since
its not clear if, for instance, two Indexes can be == if they have
different paths.

Added `ConflictIterator.__iter__` - this is required by the iterator
protocol. See
https://docs.python.org/3/library/stdtypes.html#iterator-types

### DIFF
--- a/pygit2/index.py
+++ b/pygit2/index.py
@@ -353,6 +353,22 @@ class IndexEntry:
         """The id of the referenced object as a hex string"""
         return self.id.hex
 
+    def __str__(self):
+        return "<path={} id={} mode={}>".format(self.path, self.hex, self.mode)
+
+    def __repr__(self):
+        t = type(self)
+        return "<{}.{} path={} id={} mode={}>".format(
+            t.__module__, t.__qualname__, self.path, self.hex, self.mode
+        )
+
+    def __eq__(self, other):
+        if self is other:
+            return True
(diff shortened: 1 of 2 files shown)


olsen232:
@jdavid
@rcoup FYI

jdavid:
looks good, just missing the tests

pygit2/index.py
```diff
@@ -353,6 +353,25 @@ def hex(self):
...
+        )
+
+    def __eq__(self, other):
+        if self is other:
+            return True
+        if not isinstance(other, IndexEntry):
+            return NotImplemented
+        return self.path == other.path and self.id == other.id and self.mode == other.mode
+
+    def __ne__(self, other):
```
rcoup, line 372: https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__eq__ says `__ne__` is the inverse of `__eq__` by default, so I don’t think this is needed?

jdavid: good catch

olsen232: True - I just read the README and pygit2 requires python3.6 where this is true (in python2 it is not true).

PR #1019, opened 2020-07-13
Add Repository.applies

ddevault:
To test the applicability of a patch.

`GIT_APPLY_CHECK` was added in libgit2/libgit2#5227 and shipped in 0.99.0.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -1885,9 +1885,9 @@ out:
 }
 
 PyDoc_STRVAR(Repository_apply__doc__,
-  "apply(id)\n"
+  "apply(diff or patch)\n"
   "\n"
-  "Applies the given id into HEAD.\n"
+  "Applies the given patch into HEAD.\n"
   "\n"
   "Applies a diff into HEAD, writing the results into the\n"
   "working directory.");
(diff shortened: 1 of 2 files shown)


src/repository.c
```diff
@@ -1907,6 +1907,27 @@ Repository_apply(Repository *self, PyObject *py_diff)
...
+    git_apply_location_t location = GIT_APPLY_LOCATION_WORKDIR;
+    git_apply_options options = GIT_APPLY_OPTIONS_INIT;
+    options.flags |= GIT_APPLY_CHECK;
+
+    err = git_apply(self->repo, ((Diff*)py_diff)->diff, location, &options);
+
+    if (err < 0)
+        return Py_False;
+
+    return Py_True;
```
jdavid, line 1928: Use `Py_RETURN_FALSE` and `Py_RETURN_TRUE`, otherwise it's a refcount error.

ddevault:
Thanks. It's super cool how CPython is full of silly footguns like that.

ddevault:
On a second look, I also took the opportunity to switch it to testing against the index instead of the workdir, so it can be used on bare repos (which is my use-case anyway).

jdavid:
The location should be a parameter. I would keep it simple and make the parameter mandatory (if it had a default it should be `GIT_APPLY_LOCATION_WORKDIR`, for consistency with `Repository_apply`).

jdavid:
Okay, then I will merge it.

ddevault:
Thanks!

PR #1048, opened 2020-11-13
Add support for PackBuilder methods

jeremywestwood:
A (very much WIP) attempt to add support for the Packbuilder methods from Libgit2 to create .pack and .idx files for a repository.

I just wanted to get initial input on whether the approach taken here is correct, or if I should be structuring it a different way?

I based the overall strategy on what was done for Libgit2Sharp: https://github.com/libgit2/libgit2sharp/pull/1183

**TODOs:**
* Handle the "owned" parameter correctly (if its needed?)
* Write tests

Once I've got some feedback and done the TODOs then I'll write up a proper pull request.


jdavid:
I don't think you need the `owned` parameter. Also I think you could do this only with cffi, and avoid the C code in `src/`, it should be much simpler; you can look at `pygit2/index.py` for an example of a feature (`Index`) done entirely in cffi.
Thanks

jdavid:
Looks good. Just don't see `_from_c` used anywhere, maybe you don't need it. I'll do a more detailed review when the tests will be available.

jeremywestwood:
Thanks for the comments. I've updated based on them and added the tests.

Please let me know any other feedback.

pygit2/packbuilder.py
```diff
@@ -0,0 +1,80 @@
...
+        oid = self.convert_object_to_oid(git_object)
+        err = C.git_packbuilder_insert(self._packbuilder, oid, ffi.NULL)
+        check_error(err)
+
+    def add_recur(self, git_object):
+        oid = self.convert_object_to_oid(git_object)
+        err = C.git_packbuilder_insert_recur(self._packbuilder, oid, ffi.NULL)
+        check_error(err)
+
+    def set_max_threads(self, n_threads):
```
jdavid, line 71: Docs for `git_packbuilder_set_threads` says _Set number of threads to spawn_

It doesn't say anything about **maximum**, so please rename this method to `set_threads`

pygit2/packbuilder.py
```diff
@@ -0,0 +1,80 @@
...
+        ffi.buffer(oid)[:] = git_object.raw[:]
+        return oid
+
+    def add(self, git_object):
+        oid = self.convert_object_to_oid(git_object)
+        err = C.git_packbuilder_insert(self._packbuilder, oid, ffi.NULL)
+        check_error(err)
+
+    def add_recur(self, git_object):
+        oid = self.convert_object_to_oid(git_object)
```
jdavid, line 67: You are not expecting a Git object here, but an Object id, so the argument `git_object` is misleading. I would call it `oid` but then you had to rename the local variable with the same name, maybe `git_oid`?

    def convert_object_to_oid(oid):
        git_oid = ffi.new('git_oid *')

jdavid (changes requested):
Looks good, just made a few comments.

pygit2/packbuilder.py
```diff
@@ -0,0 +1,80 @@
...
+        err = C.git_packbuilder_insert_recur(self._packbuilder, oid, ffi.NULL)
+        check_error(err)
+
+    def set_max_threads(self, n_threads):
+        return C.git_packbuilder_set_threads(self._packbuilder, n_threads)
+
+    def write(self, path):
+        err = C.git_packbuilder_write(self._packbuilder, to_bytes(path), 0, ffi.NULL, ffi.NULL)
```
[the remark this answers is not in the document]
jeremywestwood: Have updated to support NULL

pygit2/repository.py
```diff
@@ -83,6 +84,39 @@ def write(self, *args, **kwargs):
...
+        """
+
+        def pack_all_objects(pack_builder):
+            for obj in self.odb:
+                pack_builder.add(obj)
+
+        pack_delegate = pack_delegate or pack_all_objects
+
+        builder = PackBuilder(self)
+        if max_threads:
```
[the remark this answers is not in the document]
jeremywestwood: have updated to support 0 and renamed the method and argument

pygit2/packbuilder.py
jeremywestwood: I've renamed it here and the other methods which call it

jdavid:
Merged, thanks!

Issue #642, opened 2016-07-08
proxy settings


danyill:
Could pygit2 include proxy authentication settings including NTLM proxies please. That would be most useful to me. It appears to be upstream in https://github.com/libgit2/libgit2/pull/3110

sgalal:
Any updates about this?

jdavid:
PRs welcome

jdavid:
Fixed by @sathieu in PR #1063 
Open a new issue if there's something left.

PR #1063, opened 2021-02-11
Enable automatic proxy detection

sathieu:
See #642

### DIFF
--- a/pygit2/remote.py
+++ b/pygit2/remote.py
@@ -91,12 +91,24 @@ class Remote:
 
         return maybe_string(C.git_remote_pushurl(self._remote))
 
-    def connect(self, callbacks=None, direction=C.GIT_DIRECTION_FETCH):
+    def connect(self, callbacks=None, direction=C.GIT_DIRECTION_FETCH, proxy=None):
         """Connect to the remote.
+
+        Parameters:
+
+        proxy : None or True or str
+            Proxy configuration. Can be one of:
+
+            * `None` (the default) to disable proxy usage
+            * `True` to enable automatic proxy detection
+            * an url to a proxy (`http://proxy.example.org:3128/`)
         """
+        proxy_opts = ffi.new('git_proxy_options *')
+        C.git_proxy_init_options(proxy_opts, C.GIT_PROXY_OPTIONS_VERSION)
+        Remote._set_proxy(proxy_opts, proxy)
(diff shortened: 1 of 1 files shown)


pygit2/remote.py
```diff
@@ -123,6 +123,7 @@ def fetch(self, refspecs=None, message=None, callbacks=None, prune=C.GIT_FETCH_P
         with git_fetch_options(callbacks) as payload:
             opts = payload.fetch_options
             opts.prune = prune
+            opts.proxy_opts.type = C.GIT_PROXY_AUTO
```
jdavid, line 126: This should be a function argument, and the default behavior should not be modified.

Looking at the proxy options in libgit2 I think we could do an interface like this:

```
fetch(..., proxy=None) # or False (default)
fetch(..., proxy=True) # for auto
fetch(..., proxy='http://....')
```

jdavid (changes requested):

pygit2/remote.py
sathieu: @jdavid Done (with docs). Please review again.

sathieu:
I've enabled proxy configuration for 4 methods.

pygit2/remote.py
```diff
@@ -91,12 +91,30 @@ def push_url(self):
...
+            * `True` to enable automatic proxy detection
+            * an url to a proxy (`http://proxy.example.org:3128/`)
         """
+        if proxy is not None:
+            proxy_opts = ffi.new('git_proxy_options *')
+            if proxy is True:
+                proxy_opts.type = C.GIT_PROXY_AUTO
+            elif proxy is not None:
+                proxy_opts.type = C.GIT_PROXY_SPECIFIED
+                proxy_opts.url = proxy
```
jdavid, line 112: This shouldn't work because proxy will be a text string and we have to pass bytes to the C library.
You can do this: `proxy_opts.url = to_bytes(proxy)`

Also we should raise `TypeError` if `proxy`is of an unexpected type: something else than `None`, boolean or string.

And since the same code is repeated 3 times, it would be better to have a helper function, something like `set_proxy(proxy_opts, proxy)`

jdavid (changes requested):

pygit2/remote.py
sathieu: @jdavid Done. Please review again.

jdavid:
Thanks!

PR #1066, opened 2021-03-06
Use built-in functools.cached_property in py3.8+

mgorny:
Python 3.8+ supplies the cached-property functionality in functools
module, so use that instead of pulling in the backport package whenever
possible.


jdavid:
Thanks @mgorny , I've picked the first commit, but implemented a different solution for the second commit.

mgorny:
Thanks!

PR #1078, opened 2021-05-18
Add linux aarch64 wheel support

### LINKED ISSUE #1077: Add support to release aarch64 wheels
Problem
---------
On aarch64, pip install pygit2 builds the wheels from source code and then install it. It requires user to have development environment installed on his system. also, it take some time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, pip install pygit2 should download the wheels from pypi

@jdavid, Please let me know your interest on releasing aarch64 wheels. I can help in this.

odidev:
Added linux aarch64 wheel support.
Fixes #1077, @jdavid  Could you please review this PR?

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,12 +12,19 @@ jobs:
   linux:
     name: Linux
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
     steps:
     - name: Checkout pygit2
       uses: actions/checkout@v2
 
+    - name: Set up QEMU
+      if: ${{ matrix.arch == 'aarch64' }}
+      uses: docker/setup-qemu-action@v1
+
     - run: |
-        /bin/sh travis/build-all-manylinux-wheels.sh
+        /bin/sh travis/build-all-manylinux-wheels.sh ${{ matrix.arch }}
 
     - name: Save wheels
       uses: actions/upload-artifact@v2
--- a/travis/build-all-manylinux-wheels.sh
+++ b/travis/build-all-manylinux-wheels.sh
@@ -3,8 +3,14 @@
 set -x
 set -e
 
+ARCH=$1
+
 # Wait for docker pull to complete downloading container
-manylinux_image="ghcr.io/pyca/cryptography-manylinux2014:x86_64"
+if [ $ARCH = 'aarch64' ]; then
+  manylinux_image="ghcr.io/pyca/cryptography-manylinux2014_${ARCH}"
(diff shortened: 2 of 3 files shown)


.github/workflows/wheels.yml
```diff
@@ -12,12 +12,24 @@ jobs:
...
+      if: ${{ matrix.arch == 'aarch64' }}
+      uses: docker/setup-qemu-action@v1
+
+    - name: x86_64 wheels
+      if: ${{ matrix.arch == 'x86_64' }}
+      run: /bin/sh travis/build-all-manylinux-wheels.sh :x86_64
+
+    - name: aarch64 wheels
+      if: ${{ matrix.arch == 'aarch64' }}
+      run: /bin/sh travis/build-all-manylinux-wheels.sh _aarch64
```
jdavid, line 32: Please pass `matrix.arch` to `travis/build-all-manylinux-wheels.sh`, then the same step will work for both arches. The idea is to have as much logic as possible in the shell scripts, so they can be easily used from the command line, not just from CI.

For example, better to type:

```
$ sh travis/build-all-manylinux-wheels.sh aarch64
```

Than:

```
$ sh travis/build-all-manylinux-wheels.sh _aarch64
```

travis/build-all-manylinux-wheels.sh
```diff
@@ -3,8 +3,10 @@
 set -x
 set -e
 
+arch="${1}"
```
jdavid, line 6: Default to `x86_64`, so typing:

```
$ sh travis/build-all-manylinux-wheels.sh
```

Would produce the wheels for the x86_64 architecture.

Something like this may work:

```
ARCH="${1:-x86_64}"
```

jdavid (changes requested):

travis/build-all-manylinux-wheels.sh
```diff
@@ -3,8 +3,14 @@
 set -x
 set -e
 
+ARCH=$1
+
 # Wait for docker pull to complete downloading container
-manylinux_image="ghcr.io/pyca/cryptography-manylinux2014:x86_64"
+f [ $ARCH = 'aarch64' ]; then
```
jdavid, line 9: There's a typo here. It should be `if` not just `f`

jdavid (changes requested):

travis/build-all-manylinux-wheels.sh
odidev: Done.

jdavid:
Merged, thanks!

PR #1100, opened 2021-10-23
Fixes #962: First create a git_patch to get delta->flags correctly

### LINKED ISSUE #962: Wrong binary state of a delta if patches have not been iterated
Much to my surprise reading `delta.is_binary` returns the wrong value if one does not iterate over the patches beforehand. Given
```bash
mkdir test && cd test
touch README.md
dd if=/dev/urandom of=data.bin bs=256K count=1
git init .
git add README.md && git commit -m "initial"
git add data.bin && git commit -m "binary"
```
this script
```python
import pygit2

repo = pygit2.Repository('.git')
diff = repo.diff('HEAD', 'HEAD~1')

for delta in diff.deltas:
    print(delta.is_binary)

for patch in diff:
    pass

for delta in diff.deltas:
    print(delta.is_binary)
```
prints
```bash
False
True
```

yoichi:
diff_get_delta_byindex() may result incorrect value of delta->flags 
as described in the comment of git_diff_get_delta() (git2/diff.h)

```
Note that the flags on the delta related to whether it has binary
content or not may not be set if there are no attributes set for the
file and there has been no reason to load the file data at this point.
For now, if you need those flags to be up to date, your only option is
to either use `git_diff_foreach` or create a `git_patch`.
```

this pull request fixes #962

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -505,6 +505,17 @@ PyTypeObject DiffIterType = {
 PyObject *
 diff_get_delta_byindex(git_diff *diff, size_t idx)
 {
+    git_patch *patch = NULL;
+    int err;
+
+    /* create a git_patch in order to store delta->flags */
+    err = git_patch_from_diff(&patch, diff, idx);
+    if (err < 0) {
+        PyErr_SetObject(PyExc_IndexError, PyLong_FromSize_t(idx));
+        return NULL;
+    }
+    git_patch_free(patch);
+
     const git_diff_delta *delta = git_diff_get_delta(diff, idx);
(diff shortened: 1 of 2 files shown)


jdavid:
My concern is that building the patch would be a waste of time if the flags are not used later.
The alternative would be to build the patch lazily, when the flags are asked for, but this may not be easy.

I've done the related commit 464836c7 , if none of `GIT_DIFF_FLAG_BINARY` and `GIT_DIFF_FLAG_NOT_BINARY` have been set then the flags have not been loaded and the patch should be built.

PR #1101, opened 2021-10-29
feat: Wrap libgit2 git_message_trailers

thcipriani:
git-log has a formatting function to extract trailers from the commit
message field. The libgit2 function `git_message_trailers` implements
this logic and seems to catch a lot of corner cases.

Rather than ask users to extract trailers from commit messages by
fiddling with the `message` attribute of the `Commit` object, wrap the
libgit2 implementation.


thcipriani:
Hi all! This is my first attempt to generate a python module wrapping a library. I did my best to follow the patterns of the repo.

I added a test and modified the recipes doc. Let me know if I missed anything! ❤️

src/commit.c
```diff
@@ -98,6 +98,46 @@ Commit_gpg_signature__get__(Commit *self)
...
+    const char *message = git_commit_message(self->commit);
+    const char *encoding = git_commit_message_encoding(self->commit);
+
+    err = git_message_trailers(&gmt_arr, message);
+    if (err < 0)
+        return Error_set(err);
+
+    dict = PyDict_New();
+    if (dict == NULL)
+        return NULL;
```
jdavid, line 121: In this case `git_message_trailer_array_free` is not called.

src/commit.c
```diff
@@ -98,6 +98,46 @@ Commit_gpg_signature__get__(Commit *self)
...
+
+    trailer_count = gmt_arr.count;
+    for (i=0; i < trailer_count; i++) {
+        py_val = to_unicode(gmt_arr.trailers[i].value, encoding, NULL);
+        err = PyDict_SetItemString(dict, gmt_arr.trailers[i].key, py_val);
```
jdavid, line 126: `to_unicode` returns a new reference, and `PyDict_SetItemString` increments the reference count of `py_val`, so you have to decref `py_val` or there will be a reference leak

src/commit.c
jdavid: You can just `goto error;`

jdavid (changes requested):

src/commit.c
thcipriani: Good catch, thank you!

src/commit.c
thcipriani: TIL! Interesting and makes sense.

thcipriani (commented):
Thanks for the speedy review @jdavid , I fixed the errors you caught (I hope :D)

jdavid:
thanks!

PR #1109, opened 2021-11-26
Use cibuildwheel; build MacOS universal

### LINKED ISSUE #1102: Apple Silicon binary not included in current PyPI package
Running `pip install pygit2` on my M1 Pro MacBook completes without any errors. But if I then try to import it, I get the following exception:

```
aaron@aaron-wyvern ~ % python
Python 3.10.0 (default, Nov  1 2021, 21:07:02) [Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygit2
  File "<stdin>", line 1, in <module>
  File "/Users/aaron/.asdf/installs/python/3.10.0/lib/python3.10/site-packages/pygit2/__init__.py", line 30, in <module>
    from ._pygit2 import *
ImportError: dlopen(/Users/aaron/.asdf/installs/python/3.10.0/lib/python3.10/site-packages/pygit2/_pygit2.cpython-310-darwin.so, 0x0002): tried: '/Users/aaron/.asdf/installs/python/3.10.0/lib/python3.10/site-packages/pygit2/_pygit2.cpython-310-darwin.so' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/local/lib/_pygit2.cpython-310-darwin.so' (no such file), '/usr/lib/_pygit2.cpython-310-darwin.so' (no such file)
```

...with the most important bit seeming to be `incompatible architecture (have 'x86_64', need 'arm64e')`.

If I instead use `pip install pygit2 --no-binary pygit2`, the package works as expected, as long as I have `libgit2` installed on my system.

Would it please be possible to include a pre-built macOS ARM binary in the PyPI package?

simoncozens:
This does two things:

* Uses the [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) library and action to build the wheels, which simplifies the GitHub action somewhat.
* Compiles universal builds of openssl, libssh2, libgit2 and pygit2 on macOS. (Fixes #1102)


jdavid:
Great job! I didn't know about cibuildwheel

There's quite some duplication in the `build.sh`, `install-deps.sh` and `install-deps-mac.sh` scripts.
I'd prefer to have only one `build.sh`. So we will need a section:

```
if [ -n "$OPENSSL_VERSION" ]; then
    [...]
fi
```

It's okay if it only works with macOS for now.

Some other things that are now done unconditionally should be in such blocks. For example:

```
if [ "$1" = "wheel" ]; then
    [...]
else
    [...]
fi
```

May become:

```
if [ "$1" = "wheel" ]; then
    [...]
elif [ "$1" = "inplace" ]; then
    [...]
fi
```

So at the end we can have just 1 script.

simoncozens:
Ah, yes, fair comment. I will have a look at merging them back into one script on Monday.

simoncozens:
OK, I think this is done - inplace tests and CI builds both working.

jdavid:
Looks good!

Just a few comments:

- Now the artifacts are uploaded with the name `artifact` while the download step uses the name `wheels`. Whatever names used, they mast match, otherwise the `pypi` job will fail.

- The `misc/build-all-manylinux-wheels.sh` and `misc/build-manylinux-wheels.sh` are not used anymore, so they can be removed.

- Please update to cibuildwheel v2.3.0

PR #1112, opened 2021-12-11
Added Repository.merge_base_many and Repository.merge_base_octopus

frindler:
This provides an interface to git_merge_base_many and git_merge_base_octopus. One can then repeatedly merge things using merge_trees.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -565,6 +565,96 @@ Repository_merge_base(Repository *self, PyObject *args)
     return git_oid_to_python(&oid);
 }
 
+PyDoc_STRVAR(Repository_merge_base_many__doc__,
+  "merge_base_many(oids) -> Oid\n"
+  "\n"
+  "Find as good common ancestors as possible for an n-way merge.\n"
+  "Returns None if there is no merge base between the commits");
+
+PyObject *
+Repository_merge_base_many(Repository *self, PyObject *args)
+{
+    PyObject *py_commit;
+    PyObject *py_commits;
+    git_oid oid;
+    int commit_count;
+    git_oid *commits = NULL;
+    int i = 0;
+    int err;
+
+    if (!PyArg_ParseTuple(args, "O!", &PyList_Type, &py_commits))
+        return NULL;
+
+    commit_count = (int)PyList_Size(py_commits);
+    commits = malloc(commit_count * sizeof(git_oid));
+    if (commits == NULL) {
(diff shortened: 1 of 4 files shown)


jdavid (changes requested):
The body of both functions is very much the same, the only difference is one calls to `git_merge_base_many` and the other one calls to `git_merge_base_octopus`.

Instead please write a static function, so there's not duplicated code. Then it may look something like:

```c
typedef int (*git_merge_base_xxx_t)(git_oid *out, git_repository *repo, size_t length, const git_oid input_array[]);

static PyObject *
merge_base_xxx(Repository *self, PyObject *args, git_merge_base_xxx_t git_merge_base_xxx)
{
   [...]
}

PyObject *
Repository_merge_base_octopus(Repository *self, PyObject *args)
{
    return merge_base_xxx(self, args, &git_merge_base_octopus);
}
```

src/repository.c
```diff
@@ -565,6 +565,96 @@ Repository_merge_base(Repository *self, PyObject *args)
...
+
+    if (!PyArg_ParseTuple(args, "O!", &PyList_Type, &py_commits))
+        return NULL;
+
+    commit_count = (int)PyList_Size(py_commits);
+    commits = malloc(commit_count * sizeof(git_oid));
+    if (commits == NULL) {
+        PyErr_SetNone(PyExc_MemoryError);
+        return NULL;
+    }
```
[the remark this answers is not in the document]
frindler: Two good points! Fixed as you suggest.

src/repository.c
```diff
@@ -565,6 +565,79 @@ Repository_merge_base(Repository *self, PyObject *args)
...
+        py_commit_oid = PyList_GET_ITEM(py_commit_oids, i);
+        err = py_oid_to_git_oid_expand(self->repo, py_commit_oid, &commit_oids[i]);
+        if (err < 0)
+            goto out;
+    }
+
+    err = (*git_merge_base_xxx)(&oid, self->repo, commit_oid_count, (const git_oid*)commit_oids);
+
+    if (err == GIT_ENOTFOUND)
+        goto out;
```
jdavid, line 602: In this case `None` must be returned. But instead it returns `NULL` (return `NULL` means there has been an error, and then an exception must be set).

You can either use `Py_RETURN_NONE` and call `free()` before. Or incref `Py_None` and return that later.

frindler: Done!

Issue #885, opened 2019-03-04
Missing ability to "stash list"


kbd:
Hi, pygit2 is missing the ability to "stash list". The [docs](https://www.pygit2.org/working-copy.html#stash) show `stash`, `stash_apply`, `stash_drop`, and `stash_pop`, but no `stash_list`. I presume it was skipped originally because [all the stash iteration methods in libgit2](https://github.com/libgit2/libgit2/blob/master/include/git2/stash.h) require callbacks, so maybe it wasn't as straightforward to implement as the others, but it'd be nice if this function could be added.

jdavid:
This has been implemented in PR #1117 , thanks to @jorio

PR #1117, opened 2022-01-26
Add Stash type, Repository.listall_stashes

jorio:
Issue #885

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -61,6 +61,7 @@ extern PyTypeObject ReferenceType;
 extern PyTypeObject RevSpecType;
 extern PyTypeObject NoteType;
 extern PyTypeObject NoteIterType;
+extern PyTypeObject StashType;
 
 /* forward-declaration for Repsository._from_c() */
 PyTypeObject RepositoryType;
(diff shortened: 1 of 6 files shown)


src/stash.c
```diff
@@ -0,0 +1,202 @@
...
+PyTypeObject StashType;
+
+
+PyDoc_STRVAR(Stash_oid__doc__, "Stashed state.");
+
+PyObject *
+Stash_oid__get__(Stash *self)
+{
+    if (!self->oid)
+        Py_RETURN_NONE;
```
jdavid, line 45: I don't think this can be `NULL`. If it's then there was an error when it was set, and it should be handled there.
So please remove testing whether oid is `NULL` through the code. And check whether `git_oid_to_python` returns `NULL` in `foreach_stash_cb`, if it does then there has been an error.

src/stash.c
```diff
@@ -0,0 +1,202 @@
...
+}
+
+
+PyDoc_STRVAR(Stash_message__doc__, "Stash message.");
+
+PyObject *
+Stash_message__get__(Stash *self)
+{
+    if (!self->message)
+        Py_RETURN_NONE;
```
jdavid, line 58: I think the same applies here. I can imagine the message to be an empty string but not `NULL`.
So please remove testing whether message is `NULL` through the code.

src/repository.c
```diff
@@ -2211,6 +2212,59 @@ Repository_set_refdb(Repository *self, Refdb *refdb)
...
+{
+    int err;
+    Stash *py_stash;
+
+    py_stash = PyObject_New(Stash, &StashType);
+    if (py_stash == NULL)
+        return GIT_EUSER;
+
+    py_stash->oid = git_oid_to_python(stash_id);
+    py_stash->message = (message != NULL) ? strdup(message) : NULL;
```
jdavid, line 2225: As said in the other comments, I think `message` and `stash_id` cannot be `NULL`. So please add an `assert` statement for each one. Then check whether `git_oid_to_python` or `strdup` return `NULL`, that would mean a memory error.

src/stash.c
```diff
@@ -0,0 +1,202 @@
...
+#include "object.h"
+#include "error.h"
+#include "types.h"
+#include "utils.h"
+#include "oid.h"
+
+PyTypeObject StashType;
+
+
+PyDoc_STRVAR(Stash_oid__doc__, "Stashed state.");
```
jdavid, line 39: Please change to the more informative `The commit id of the stashed state.`

src/stash.c
```diff
@@ -0,0 +1,202 @@
...
+
+PyObject *
+Stash_oid__get__(Stash *self)
```
jdavid, line 42: Please rename to `commit_id`, I think that's more clear.

jdavid (changes requested):

jorio:
Thank you for your review! I have addressed your feedback in 1205eb7 and added some documentation for the new features as well.

jdavid:
Merged, thanks for your contribution!

PR #1118, opened 2022-01-26
Clean up obsolete function declarations in C headers

jorio:
As I was preparing my previous PR, I noticed that many functions declared in C headers aren't implemented or used anywhere.

I suspect these declarations are remnants of previous refactors.

After removing the obsolete declarations, some header files were empty so I deleted those altogether.


jdavid:
Good catch

Issue #647, opened 2016-08-11
GPG-signed commits


parisk:
Is there any way to sign commits (and tags maybe) with pygit2? I performed a quick search about `gpg` and `signed` both in the repository and documentation website and could not find any related document.

parisk:
Sorry for nagging here. Can any of the maintainers share if this can happen right now, if there is any kind of workaround for this or, there are any features that need to be implemented first?

Thanks 🙂 .

webknjaz:
I assume that it may be possible by manually assigning [`Commit.gpg_signature`](https://www.pygit2.org/objects.html#Commit.gpg_signature).

cc @jdavid

webknjaz:
Looks like I was wrong. pygit2 only exposes already existing signatures.

But I've collected some pointers:
* It seems like more code need to be added around https://github.com/libgit2/pygit2/blob/0462c5ca1c9147804e15d594aafc086d7e4e3f98/src/repository.c#L1001 to optionally sign commits
* Producing a signed commit will probably involve invoking [`git_commit_create_buffer`](https://libgit2.org/libgit2/#HEAD/group/commit/git_commit_create_buffer) to create an in-memory commit in a [`git_buf`](https://libgit2.org/libgit2/#HEAD/type/git_buf) object having unsigned `commit_content` that will be passed to [`git_commit_create_with_signature`](https://libgit2.org/libgit2/#HEAD/group/commit/git_commit_create_with_signature). After this is done, I assume we'll need to update the ref manually because unlike `git_commit_create`, `git_commit_create_with_signature` does not have `update_ref`.
* Also we need to set a signing callback in [`git_rebase_options`](https://libgit2.org/libgit2/#HEAD/type/git_rebase_options): https://github.com/libgit2/libgit2/blob/b0692d6b3e818b9389295d7d33a0601143cc0c16/include/git2/rebase.h#L84 / [`git_commit_signing_cb`](https://libgit2.org/libgit2/#HEAD/group/callback/git_commit_signing_cb)

I'd expect the public API of pygit2 to just accept some args via [`Repository.create_commit()`](https://www.pygit2.org/objects.html#pygit2.Repository.create_commit) toggling the signing as well as maybe some global settings/callbacks for defaults.

kuwv:
I really think is a really important feature and in my opinion should be prioritized. Without it it's too easy to impersonate people even on GitHub. Without this capability users are forced between choosing security or automation but not both.

kuwv:
[...]

PR #1142, opened 2022-05-02
feat: add commit signing

kuwv:
*Overview*

This pull request will add commit signing capability to the repository object. It is modeled off both the libgit2 and rugged implementations. It will allow integration with a 3rd party GPG signing tool.

*References*

https://github.com/libgit2/pygit2/issues/647
https://github.com/libgit2/rugged/blob/master/ext/rugged/rugged_commit.c#L791
https://github.com/libgit2/rugged/blob/master/ext/rugged/rugged_commit.c#L842
https://github.com/libgit2/pygit2/blob/master/test/test_commit_gpg.py


kuwv:
@jdavid I am currently writing the tests for this but I wanted to get this in front of you first.

jdavid:
Thanks @kuwv, looks good.
There is some code duplication, but it doesn't look easy to avoid, so it's ok.
Only waiting for the tests then.

kuwv:
> Thanks @kuwv, looks good. There is some code duplication, but it doesn't look easy to avoid, so it's ok. Only waiting for the tests then.

Yeah, I guess two functions could potentially be combined later but I didn't want to introduce any breaking changes to the API. So, I choose to model Rugged and git2-rs.

One question, together neither function updates the HEAD reference. The Rugged implementation also does not. Should I add the reference update to commit_with_signature or should the API user just do it?

jdavid:
Please use `gpgsigned.zip` and remove `gpgsigned.git`

In my latest commit b472bdaff I've moved to use `.zip` files only, because with libgit2 1.4.3 there were ownership issues with `.tar` and raw `.git` data. I'm surprised that the tests pass for aarch64, but even then I think it's better to only use zip files.

kuwv:
> Please use `gpgsigned.zip` and remove `gpgsigned.git`

sure 

> I've moved to use `.zip` files only...

Yup, I saw that. Pardon my cruft, just needed to check.

Aside from that should I be worried about the appveyor tests?

kuwv:
Alright, should be good now.

jdavid:
> Aside from that should I be worried about the appveyor tests?

No, the appveyor tests are failing because of the upgrade to libgit2 v1.4.3, so it's unrelated.

> Alright, should be good now. Any additional changes you would like to see here?

Only one last thing: can you rebase and squash?
At least commits e0880ce8739cb2f18fd96c36b2d387be009f5590 and c8e55097456c0c59cfd8eb619e4e6db8a6657422 should be squashed so we don't to keep `.gpgsigned.git` in history, which would add weight to the repo.

> Maybe shorten create_commit_with_signature to create_commit_signature or add_commit_signature?

No, I think the current name is fine, we often follow libgit2's naming.

kuwv:
@jdavid Should be good now.

jdavid:
Thanks!

PR #1151, opened 2022-07-19
status: add untracked_files and ignored kwargs

dtrifiro:
add `ignored` and `untracked_files` kwargs to `status`.

`untracked_files` can take be one of `"no"`, `"normal"`, `"all"`, which return respectively: no untracked files, untracked files/directories without recursing into subdirectories, all untracked files/directories.
`ignored` can be `True` (show ignored files) or `False` (do not show ignored files). When`untracked_files="no"`, this does not have any effect.

For more information see `--untracked-files` and `--ignored`, see https://book.git-scm.com/docs/git-status


pygit2/_pygit2.pyi
```diff
@@ -496,7 +496,7 @@ class Repository:
     def revparse_single(self, revision: str) -> Object: ...
     def set_odb(self, odb: Odb) -> None: ...
     def set_refdb(self, refdb: Refdb) -> None: ...
-    def status(self) -> dict[str,int]: ...
+    def status(self, untracked_files: str = "all", ignored: bool = True) -> dict[str,int]: ...
```
dtrifiro, line 499: In https://book.git-scm.com/docs/git-status `--ignored` is actually a string and can take three values:
- "traditional": Shows ignored files and directories, unless --untracked-files=all is specified, in which case individual files in ignored directories are displayed.
-  "no": Show no ignored files
- "match": Shows ignored files and directories matching an ignore pattern.

I'd rather avoid implementing `match`, but I could make `ignored` a `str` kwarg, raising `NotImplementedError` for `match`

test/test_status.py
```diff
@@ -32,3 +36,49 @@ def test_status(dirtyrepo):
...
+            {
+                "new_file",
+                os.path.join("subdir", "new_file"),
+                "staged_delete_file_modified",
+                os.path.join("untracked_dir", "untracked_file"),
+            },
+        ),
+    ],
+)
+def test_status_untracked_normal(dirtyrepostatus, untracked_files, expected):
```
dtrifiro, line 70: I added a new `.zip` archive which is a copy of `dirtyrepo` which also includes a `.gitignore` and an ignored file.

test/test_status.py
```diff
@@ -32,3 +36,49 @@ def test_status(dirtyrepo):
...
         assert filepath in git_status
         assert status == git_status[filepath]
+
+
+def test_status_untracked_no(dirtyrepo):
+    git_status = dirtyrepo.status(untracked_files="no")
+    not any(status & pygit2.GIT_STATUS_WT_NEW for status in git_status.values())
+
+
+@pytest.mark.parametrize(
```
dtrifiro, line 46: These tests don't actually cover all of the possible scenarios, such as having tracked subdirectories with ignored files, tracked subdirectories with untracked files, untracked subdirectories with ignored files, etc, etc. Might improve these tests soon.

test/conftest.py
```diff
@@ -48,6 +48,12 @@ def dirtyrepo(tmp_path):
         yield pygit2.Repository(path)
 
 
+@pytest.fixture
+def dirtyrepostatus(tmp_path):
+    with utils.TemporaryRepository('dirtyrepostatus.zip', tmp_path) as path:
```
dtrifiro, line 53: `dirtyrepostatus.zip` is copy of `dirtyrepo.zip` which also includes a `.gitignore` and an ignored file.

dtrifiro:
Not sure why appveyor failed, one of the errors I'm getting:
```failed on setup with "_pygit2.GitError: C:\Users\appveyor\AppData\Local\Temp\1\pytest-of-appveyor\pytest-0\test_reference_set_sha0\testrepo: repository path 'C:/Users/appveyor/AppData/Local/Temp/1/pytest-of-appveyor/pytest-0/test_reference_set_sha0/testrepo/' is not owned by current user"
```

As I'm not able to reproduce this locally, I could use some help there

pygit2/_pygit2.pyi
dtrifiro: I have no strong opinion here, so I'm open to whatever you prefer here

skshetry:
> Not sure why appveyor failed, one of the errors I'm getting:
> 
> > failed on setup with "_pygit2.GitError: C:\Users\appveyor\AppData\Local\Temp\1\pytest-of-appveyor\pytest-> 0\test_reference_set_sha0\testrepo: repository path 'C:/Users/appveyor/AppData/Local/Temp/1/pytest-of-appveyor/pytest-0/test_reference_set_sha0/testrepo/' is not owned by current user"
> 
> As I'm not able to reproduce this locally, I could use some help there

This is also failing on `master`: https://ci.appveyor.com/project/jdavid/pygit2/builds/44206064/job/83rhjdo521x5uk6e.
#1150 seems to be the relevant PR where it's also failing, due to `GIT_OPT_GET_OWNER_VALIDATION` being enabled by default.

jdavid:
> This is also failing on master [...]

This is now fixed in master, you can rebase to get AppVeyor green.

pygit2/_pygit2.pyi
jdavid: It's ok as bool

test/test_status.py
jdavid: We're just passing the options to libgit2, the purpose of the test suite is not to test libgit2, which already has its own test suite. So better to keep the tests simple, it's good the way they are already.

test/conftest.py
jdavid: If the new `dirtyrepostatus.zip` actually replaces `dirtyrepo.zip` will the test suite pass? If it does then better to keep only one zip (for the repo size to be smaller better than bigger).

src/repository.c
```diff
@@ -1771,20 +1772,64 @@ Repository_compress_references(Repository *self)
...
+
+    if (!PyArg_ParseTupleAndKeywords(args, kw, "|sO", kwlist,
+                                     &untracked_files, &ignored))
+        goto error;
+
+    git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+    opts.flags = GIT_STATUS_OPT_DEFAULTS;
+
+    if (!strcmp(untracked_files, "no")){
+       opts.flags &= ~(GIT_STATUS_OPT_INCLUDE_UNTRACKED | GIT_STATUS_OPT_RECURSE_UNTRACKED_DIRS);
```
jdavid, line 1814: (Style) There is one trailing white space in this line.

src/repository.c
```diff
@@ -1771,20 +1772,64 @@ Repository_compress_references(Repository *self)
...
+    PyObject* ignored = Py_False;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kw, "|sO", kwlist,
+                                     &untracked_files, &ignored))
+        goto error;
+
+    git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+    opts.flags = GIT_STATUS_OPT_DEFAULTS;
+
+    if (!strcmp(untracked_files, "no")){
```
jdavid, line 1813: (Style) Please add and space between the `)` and the `{`
Same thing in other lines.

src/repository.c
```diff
@@ -1771,20 +1772,64 @@ Repository_compress_references(Repository *self)
...
+    } else if (strcmp(untracked_files, "all")){
+        return PyErr_Format(
+          PyExc_ValueError,
+          "untracked_files must be one of \"all\", \"normal\" or \"one\"");
+    };
+
+    if (!PyBool_Check(ignored)){
+        return PyErr_Format(
+          PyExc_TypeError,
+          "ignored must be True or False");
```
jdavid, line 1826: (Style) If you put these 3 lines in 1, it won't bee too wide.

test/test_status.py
```diff
@@ -32,3 +36,49 @@ def test_status(dirtyrepo):
...
+    "untracked_files,expected",
+    [
+        ("no", set()),
+        (
+            "normal",
+            {
+                os.path.join("untracked_dir", ""),
```
jdavid, line 53: Use `pathlib.Path` instead of `os.path`
Same in other places.

jdavid: Actually this probably must be different, since this test is failing on Windows (AppVeyor) because of different path separator.
Probably it must simply be `untracked_dir/`, since Git returns Unix-style path separators.

jdavid (changes requested):

test/conftest.py
dtrifiro: Didn't modify `dirtyrepo.zip` in place because I was afraid it might break tests. Will try running the test suite replacing `dirtyrepo.zip` and wil keep it that way if no tests break.

dtrifiro: It's kinda tricky to update `dirtyrepo.zip` without breaking other tests. The easiest solution seems to be adding a `.gitignore` to `dirtyrepo` (without committing it) and then modifying these two tests so that they pass:

```diff --git a/test/test_diff.py b/test/test_diff.py
index 55269ca..61007c6 100644
--- a/test/test_diff.py
+++ b/test/test_diff.py
@@ -64,6 +64,7 @@ index 297efb8..0000000
 PATCHID = 'f31412498a17e6c3fbc635f2c5f9aa3ef4c1a9b7'
 
 DIFF_HEAD_TO_INDEX_EXPECTED = [
+    '.gitignore',
     'staged_changes',
     'staged_changes_file_deleted',
     'staged_changes_file_modified',
@@ -87,6 +88,7 @@ DIFF_HEAD_TO_WORKDIR_EXPECTED = [
 ]
 
 DIFF_INDEX_TO_WORK_EXPECTED = [
+    '.gitignore',
     'file_deleted',
     'modified_file',
     'staged_changes_file_deleted',
```

Would this be ok?

jdavid: Yes

dtrifiro:
Thanks for the feedback! Updated everything as discussed

jdavid (approved):

jdavid:
Merged, thanks!

PR #1155, opened 2022-08-14
Fix Signature repr()

JacobSwanson:
`to_unicode()` expects a non-NULL `x` argument (due to `strlen(x)`);
however `self->encoding` _may_ be NULL, especially if built with
`build_signature()` (i.e. if called from the `commit.author` property).

As a result, any `repr()` on a `Signature` with a NULL `encoding` will
cause a segfault.

Other functions manually map NULL to "utf-8" (see
`Signature__encoding__get__()`), so we'll do the same thing here.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -45,12 +45,13 @@ Signature_init(Signature *self, PyObject *args, PyObject *kwds)
     int offset = 0;
 
     if (!PyArg_ParseTupleAndKeywords(
-            args, kwds, "Os|Lis", keywords,
+            args, kwds, "Os|Liz", keywords,
             &py_name, &email, &time, &offset, &encoding))
         return -1;
 
     PyObject *tname;
-    const char *name = pgit_borrow_encoding(py_name, encoding, NULL, &tname);
+    const char *name = pgit_borrow_encoding(
+        py_name, value_or_default(encoding, "utf-8"), NULL, &tname);
     if (name == NULL)
         return -1;
 
@@ -104,13 +105,10 @@ PyDoc_STRVAR(Signature__encoding__doc__, "Encoding.");
 PyObject *
 Signature__encoding__get__(Signature *self)
 {
-    const char *encoding;
+    if (self->encoding)
+        return to_encoding(self->encoding);
 
-    encoding = self->encoding;
-    if (encoding == NULL)
-        encoding = "utf-8";
-
-    return to_encoding(encoding);
+    Py_RETURN_NONE;
 }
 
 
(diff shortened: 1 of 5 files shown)


jdavid:
Good catch.

I think it would be more correct to represent `NULL` as `None`, so in this case the repr looks like:

```python
pygit2.Signature('Foo Ibáñez', 'foo@bar.com', 1322174594, 60, None)
```

Then for this to work `Signature_init` has to be changed to use `z` instead of `s`:

```C
    if (!PyArg_ParseTupleAndKeywords(
            args, kwds, "Os|Liz", keywords,
            &py_name, &email, &time, &offset, &encoding))
```

JacobSwanson:
Agreed that representation has better semantics.  Updated to reflect!

src/signature.c
```diff
@@ -137,7 +135,10 @@ PyDoc_STRVAR(Signature_name__doc__, "Name.");
 PyObject *
 Signature_name__get__(Signature *self)
 {
-    return to_unicode(self->signature->name, self->encoding, NULL);
+    return to_unicode(
+        self->signature->name,
+        value_or_default(self->encoding, "utf-8"),
+        NULL);
```
jdavid, line 141: This change is not needed, since `to_unicode` already handles the case when the given `encoding` is `NULL`.

src/signature.c
```diff
@@ -146,7 +147,10 @@ PyDoc_STRVAR(Signature_email__doc__, "Email address.");
 PyObject *
 Signature_email__get__(Signature *self)
 {
-    return to_unicode(self->signature->email, self->encoding, NULL);
+    return to_unicode(
+        self->signature->email,
+        value_or_default(self->encoding, "utf-8"),
+        NULL);
```
jdavid, line 153: idem as above

test/test_signature.py
```diff
@@ -54,6 +54,14 @@ def test_latin1():
     assert signature.email == signature.raw_email.decode(encoding)
     assert signature.email.encode(encoding) == signature.raw_email
 
+def test_none():
+    signature = Signature('Foo Ibáñez', 'foo@example.com', encoding=None)
+    assert None == signature._encoding
```
jdavid, line 59: Comparison with `None` is better done with the `is` operator.

jdavid (changes requested):
Also, please rebase, and squash into a single commit.

JacobSwanson:
Fixed up! Let me know if there is anything else I can do.  Thanks

jdavid:
One more thing, `pygit2/_pygit2.pyi` needs to be updated. I think only `_encoding` and `__init__` are concerned:

```python
class Signature:
    _encoding: str | None
    ...
    def __init__(self, name: str, email: str, time: int, offset: int, encoding: str | None) -> None: ...
```

jdavid:
Merged, thanks!

jdavid:
FYI, after a second thought I've done follow up commit. The main change is that ``Signature._encoding`` will return ``utf-8`` when internally it's ``NULL``. This is to keep backwards compatibility, so users don't need to worry about handling ``None``.

PR #1159, opened 2022-09-25
Add RemoteCollection.names() iterator

agausmann:
# Motivation

This `names()` iterator provides a big benefit in performance in my use case, where I want to test whether or not a remote exists for a large number of potential remote names (hundreds)

The first code I used looked something like:
```python
try:
    repo.remotes[remote_name]
except KeyError:
    repo.remotes.create(remote_name, remote_url)
```
However, this performs a lookup for all remote names, which was prohibitively expensive. So I decided to try collecting all the names into a set beforehand:
```python
existing_remotes = set(remote.name for remote in repo.remotes)

for remote_name, remote_url in wanted_remotes:
    if remote_name not in existing_remotes:
        repo.remotes.create(remote_name, remote_url)
```
However, this still performs a lookup for each remote to create the `Remote` objects yielded by the iterator.

With `RemoteCollection.names()`, I can construct the set much more efficiently:
```python
existing_remotes = set(repo.remotes.names())
```

### DIFF
--- a/pygit2/remote.py
+++ b/pygit2/remote.py
@@ -295,20 +295,12 @@ class RemoteCollection:
             C.git_strarray_free(names)
 
     def __iter__(self):
-        names = ffi.new('git_strarray *')
-
-        try:
-            err = C.git_remote_list(names, self._repo._repo)
+        cremote = ffi.new('git_remote **')
+        for name in self._ffi_names():
+            err = C.git_remote_lookup(cremote, self._repo._repo, name)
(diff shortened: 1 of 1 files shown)


jdavid:
Thanks, please add a unit test.

jdavid:
For reference I think iteration should yield the names, like references/branches do, but that breaks compatibility.
Also I'm unsure which is best `names()` or `keys()`, names is more like git, keys is more like python.

Issue #1161, opened 2022-10-01
Fatal Python error: none_dealloc: deallocating None when getting commit author


siscia:
Hello everybody, and thanks for maintaining this repository.

I am reporting a runtime error `Fatal Python error: none_dealloc: deallocating None`

I am iterating over commits of a repository and getting information about the commiter:

```
import pygit2 # type: ignore

# this is a git clone of
# git@github.com:facebook/react.git
repo_path = "tests/react"

repo = pygit2.Repository(repo_path)

commits = repo.walk(repo.head.target, pygit2.GIT_SORT_REVERSE)

# we are missing what is happening at the first commit
for commit in commits:
    author = commit.author
    print(f"{commit.id=}")
    print(f"{author=}")
```

The execution terminates with:

```
... more logs line...
commit.id=eb056fcf5d12f100e4f8df700b634d3c4cc5f9ba
author=pygit2.Signature('Paul O’Shannessy', 'paul@oshannessy.com', 1439962431, 4294966876, None)
commit.id=ec9b866bc0fa4118f0b3750589e26ed851ef97bd
author=pygit2.Signature('Paul O’Shannessy', 'paul@oshannessy.com', 1439946313, 4294966876, None)
commit.id=50adfeb4217a6086fe7d9c4be5d9db732614106e
author=pygit2.Signature('Paul O’Shannessy', 'paul@oshannessy.com', 1440001562, 4294966876, None)
commit.id=1e10bb5923b32ab3b6c86c65c7ed18f693056da5
Fatal Python error: none_dealloc: deallocating None
Python runtime state: initialized

Current thread 0x00007f423d6f21c0 (most recent call first):
  File "/home/simo/projects/codedigger/repro.py", line 14 in <module>

Extension modules: pygit2._pygit2, _cffi_backend (total: 2)
Aborted (core dumped)
```

It is not deterministic when it terminates, and it always terminates at a different commit.

siscia:
It is not super clear to me this call https://github.com/libgit2/pygit2/blob/efa451d51f2c0f6f2c14f190b1a8781992ec7177/src/signature.c#L248

As I would expect that we need to increase the reference counting for that none.

siscia:
Yep, it was that!

Preparing PR

PR #1162, opened 2022-10-03
Remove decrement of reference when using Py_None #1161

siscia:
Addressing #1161 

If the representation of a signature requires the use of `Py_None` we should not decrement the reference count of the special `Py_None` object.

I tried adding a unit test to verify this new behavior, unfortunately, all the unit test I could come up with were not triggering the bug.

This was tested against the reproduce example in the issue setting the env var PYTHONPATH.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -202,3 +202,4 @@ Authors::
   chengyuhang
   earl
   odidev
+  siscia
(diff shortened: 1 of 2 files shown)


jdavid:
Good catch!
The correct solution is to `Py_INCREF(Py_None)` when it is assigned. That's what we do in other places, and what is done in Python's core. Please update the PR.

siscia:
Wonderful I will try to follow up tonight!

siscia:
Addressed!

Let me know if you need me to squash the commits or to fix the history somehow!

jdavid:
Yes please, squash the commits.

siscia:
Done!

PR #1183, opened 2022-12-08
Improve write_archive performance

xMez:
Accessing each item in the tree during this loop slows down the `write_archive` method significantly when dealing with repositories with a large nr of files.

On a repository of ~5000 files the archiving time went from `13s` to `0.5s`.  
From what I can tell when testing `tree[entry.path].filemode` and `entry.mode` are the same.


jdavid:
Thanks

Issue #1189, opened 2023-02-13
Missing binding for libgit2 git_repository_hashfile


muhqu:
While there is [`pygit2.hashfile`](https://github.com/libgit2/pygit2/blob/fce010dc5fa590f0d832b942bdd1d60ea904f89a/src/pygit2.c#L127-L152) which is a binding for libgit2 [`git_odb_hashfile`](https://libgit2.org/libgit2/#v1.5.1/group/odb/git_odb_hashfile), there appears to be no binding for [`git_repository_hashfile`](https://libgit2.org/libgit2/#v1.5.1/group/repository/git_repository_hashfile).

The difference is that `git_repository_hashfile` calculates the file hash using repository filtering rules (e.g. crlf filters), while `git_odb_hashfile` doesn't apply any filtering rules.

Would it be possible to add `hashfile` method to `pygit2.Repository` which maps to `git_repository_hashfile`?

muhqu:
@jdavid Thanks for merging my PR #1190  🙏 

…do you plan to make a release to ship this change in the coming days?

muhqu:
Also … any idea why `git_repository_hashfile` is implemented to error when `core.safecrlf` is `true` and a files line endings are not correct? …see [libgit2 API docs](https://libgit2.org/libgit2/#v1.5.1/group/repository/git_repository_hashfile):
> […] if the repository has core.safecrlf set to fail and the filtering triggers that failure, then this function will return an error and not calculate the hash of the file.

I'm just wondering, as the git command `git hash-object` does not behave the same in this case, it will always return a hash while taking the crlf filtering into account…

jdavid:
No idea, that's probably a question for the libgit2 developers.

About the release I first want to at least review PR #1191

jdavid:
I was going to make the release when I saw that tests fail in AppVeyor. It's the tests of the hashfile function, could you have a look? https://ci.appveyor.com/project/jdavid/pygit2/builds/46329622/job/iv77sk6d184mg3ep

muhqu:
@jdavid Oh no, I'm sorry for that... I've spent some time today to looks into it, but I have not built on windows for quite some time and so I'm still struggling to get a successful build running... 😞 

I'm totally fine if you simply revert my merged PR to get the release out. And we can look into the issue in a new PR that again reverts the revert... _does that make sense?_ 🤔    https://github.com/libgit2/pygit2/pull/1194

muhqu:
[...]

PR #1190, opened 2023-02-13
Add Repository.hashfile(path) which maps libgit2 git_repository_hashfile

muhqu:
Might serve as a fix for #1189.

* Added APIs
  * `Repository.hashfile(path: str) -> Oid`
* Added unit tests
  * `test/test_repository.py::test_repo_hashfile_same_hash`
  * `test/test_repository.py::test_repo_hashfile_crlf_normalization`
  * `test/test_repository.py::test_repo_hashfile_no_normalization`
  * `test/test_repository.py::test_repo_hashfile_crlf_normalization_error`


src/repository.c
```diff
@@ -2435,6 +2435,33 @@ Repository_listall_stashes(Repository *self, PyObject *args)
     }
 }
 
+PyDoc_STRVAR(Repository_hashfile__doc__,
+  "hashfile(path: str) -> Oid\n"
+  "\n"
+  "Calculate hash of file using repository filtering rules.");
```
muhqu, line 2441: We might want to add the following note:
> If you simply want to calculate the hash of a file on disk with no filters, you can just use the `pygit2.hashfile()` API. However, if you want to hash a file in the repository and you want to apply filtering rules (e.g. crlf filters) before generating the SHA, then use this function.
> 
> Note: if the repository has `core.safecrlf` set to fail and the filtering triggers that failure, then this function will return an error and not calculate the hash of the file.

PR #1191, opened 2023-02-13
Add references iterator

sbhuller98:
This creates the ability to use a references iterator. You can also filter the type of reference you want without first having to load it into memory.


pygit2/repository.py
```diff
@@ -291,6 +291,33 @@ def create_reference(self, name, target, force=False, message=None):
         return self.create_reference_symbolic(name, target, force,
                                               message=message)
 
+    def refs_iterator(self, references_return_type = 0):
```
jdavid, line 294: Please move this to `Repository.references.iterator(...)`.

jdavid (changes requested):

src/repository.c
```diff
@@ -1406,6 +1407,100 @@ Repository_listall_reference_objects(Repository *self, PyObject *args)
...
+  "Creates and returns an iterator for references.");
+
+PyObject *
+Repository_references_iterator_init(Repository *self, PyObject *args)
+{
+    int err;
+    git_reference_iterator *iter;
+    RefsIterator *refs_iter;
+
+    refs_iter = PyObject_New(RefsIterator, &RefsIteratorType);
```
jdavid, line 1438: And check whether `refs_iter` is `NULL`

src/repository.c
```diff
@@ -1406,6 +1407,102 @@ Repository_listall_reference_objects(Repository *self, PyObject *args)
...
+    return (PyObject*)refs_iter;
+}
+
+static PyObject *
+wrap_and_check_reference(Repository *self, git_reference *ref) {
+    PyObject *py_ref = wrap_reference(ref, self);
+    if (py_ref == NULL) {
+        Py_INCREF(Py_None);
+        return Py_None;
+    }
```
jdavid, line 1456: This is masking an error, if wrap_reference returns NULL it's because there was an error, and it must be forwarded.

src/repository.h
```diff
@@ -33,6 +33,12 @@
 #include <git2.h>
 #include "types.h"
 
+typedef enum {
+	GIT_REFERENCES_ALL = 0,
+	GIT_REFERENCES_BRANCHES = 1,
+	GIT_REFERENCES_TAGS = 2,
+} git_reference_iterator_return_t;
```
jdavid, line 40: Linking with my previous comment, probably these should be exposed to Python with `ADD_CONSTANT_INT` and used both by `Repository.references.iterator(...)` and by the `Repository_references_iterator_next` function.

jdavid (changes requested):

pygit2/repository.py
sbhuller98: I also updated Repository.references.__iter__(). Let me know if that is not ok

pygit2/repository.py
```diff
@@ -1549,8 +1550,42 @@ def get(self, key: str):
...
     def __iter__(self):
-        for ref_name in self._repository.listall_references():
-            yield ref_name
+        iter = self._repository.references_iterator_init()
+        while True:
+            ref = self._repository.references_iterator_next(iter)
+            if ref:
+                yield ref.name
+            else:
+                return
```
jdavid, line 1559: I think this could be written as `yield from self.iterator()`
Or even simpler, make them the same function `__iter__ = iterator`

jdavid (changes requested):

pygit2/repository.py
jdavid: Forget that one, I read it too fast.

jdavid:
Thanks for your contribution!

PR #1204, opened 2023-03-28
Add Repository.state

jorio:
It's already possible to clear the state of the repository (state_cleanup()), but there was no way to retrieve that state (cherrypicking, merging, etc.).


jdavid:
I've merged before you rebased. Thanks for another contribution!

PR #1229, opened 2023-07-25
remotes: support anonymous remotes

pmrowla:
Adds support for `git_remote_create_anonymous` for creating anonymous (in-memory) remotes when you have a URL but do not want to create a named remote and the associated config entries

### DIFF
--- a/pygit2/remote.py
+++ b/pygit2/remote.py
@@ -344,6 +344,16 @@ class RemoteCollection:
 
         return Remote(self._repo, cremote[0])
 
+    def create_anonymous(self, url):
+        """Create a new anonymous (in-memory only) remote with the given URL.
+        Returns a <Remote> object.
+        """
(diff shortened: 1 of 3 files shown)


pygit2/remote.py
```diff
@@ -344,6 +344,16 @@ def create(self, name, url, fetch=None):
 
         return Remote(self._repo, cremote[0])
 
+    def create_anonymous(self, url):
+        """Create a new anonymous (in-memory only) remote with the given URL.
+        Returns a <Remote> object.
+        """
```
pmrowla, line 350: This could also be done in the existing `create()` by checking something like
```
if not name:
    C.git_remote_create_anonymous
```
but I wasn't sure if that would be explicit enough since `create(None, url)` could also imply that the user intended to make a libgit2 `detached` remote instead of an `anonymous` one.

PR #1237, opened 2023-10-02
filter: add API for registering Python filters

pmrowla:
- Adds `pygit2.filter` API for registering pure-python `git_filter_*` filters.
- Adds `pygit2.BlobIO()` wrapper for reading raw and filtered blob contents.
    - `BlobIO` will stream buffered content (instead of loading the entire blob into memory) if the underlying libgit2 filter(s) support the streaming filter API.
- Adds `Repository.get_attr(..., commit=...)` to support attr lookup by commit ID
    - The commit specific attrs are already provided to filters in `Filter.check()` but it is useful for anyone implementing filters to lookup commit specific attributes before actually running the filter (i.e. to pre-fetch LFS objects in a batch rather than streaming them one by one upon checkout)

This allows the user to subclass `pygit2.Filter` and then register their own Python smudge/clean filters.

In libgit2, you normally need to define both a `git_filter` and a `git_writestream` implementation for your filter. This PR abstracts this out into a single `pygit2.Filter` class so the user does not need to work directly with the `git_writestream` linked list chain that is invoked when libgit2 does filtering.


pygit2/filter.py
```diff
@@ -0,0 +1,461 @@
...
+    filter should only write output data in `close()`.
+    """
+
+    # Space-separated string list of attributes to be used in `check()`
+    attributes = ""
+
+    def __init__(self):
+        pass
+
+    def check(self, src: FilterSource, attr_values: List[str]):
```
pmrowla, line 95: `Filter.check()` is equivalent to `git_filter->check()`

pygit2/filter.py
```diff
@@ -0,0 +1,461 @@
...
+        If `Passthrough` is raised, the filter will not be applied.
+        """
+        pass
+
+    def write(
+        self,
+        data: bytes,
+        src: FilterSource,
+        write_next: Callable[[bytes], None]
+    ):
```
pmrowla, line 110: `Filter.write()` is equivalent to `git_writestream->write()` (i.e `git_filter->stream->write`).

`write_next` writes to the next stream in the linked-list of `git_writestream`'s (where there is one stream for each filter in the chain)

pygit2/filter.py
```diff
@@ -0,0 +1,461 @@
...
+        `write()` may be called multiple times per stream.
+
+        Output data should be written to `write_next` whenever it is available.
+        """
+        write_next(data)
+
+    def close(
+        self,
+        write_next: Callable[[bytes], None]
+    ):
```
pmrowla, line 123: `Filter.close()` is equivalent to `git_writestream->close()` (i.e. `git_filter->stream->close()`)

jdavid (changes requested):

jdavid:
Hi @pmrowla , thanks for the contribution.
Most important is that tests don't pass, apparently it gets stuck in `test_blob_filter` wit pypy

pmrowla:
Thanks for the review, I'll make the suggested changes and take a look into the pypy issue.

I'll likely keep the PR marked as draft for now, I'm still testing to make sure this works with real-world LFS filtering (for LFS) so it's still possible this PR may change a bit.

src/blob.c
```diff
@@ -137,9 +139,131 @@ Blob_diff_to_buffer(Blob *self, PyObject *args, PyObject *kwds)
...
+  "\n"
+  "as_path : str\n"
+  "    Treat blob as if it had this filename (used for attribute lookups).\n"
+  "\n"
+  "flags\n"
+  "    GIT_BLOB_FILTER_* bitflags.\n"
+  "\n"
+  "commit_id: oid\n"
+  "    Commit to load attributes from when\n"
+  "    GIT_BLOB_FILTER_ATTRIBUTES_FROM_COMMIT is specified.\n");
```
pmrowla, line 199: Updated this to write to a file-like object instead of returning a bytestring.

With filtered content it's more likely that the resulting data will be large (especially in the LFS use case) and we may not want to or be able to load the entire object into memory (which is what `git_blob_filter()` does).

Using a stream object here allows us to do buffered writes (and the user can still get the full bytestring in memory if they want via `io.BytesIO`)

pmrowla: I think I will probably end up reworking this again so that it returns something like an `io.BufferedReader` instance where the caller can `read()` when they want to instead of needing to write the full filtered blob into an intermediate file first

pmrowla:
After some investigation, it looks like pypy and cffi don't always mix well if an ffi callback is run outside of the main thread, which is what was causing the pypy tests to hang (since the libgit2 filtering can be done in a separate thread).

I've rewritten the underlying filter/writestream code to done entirely in the _pygit2 C-extension (so this PR no longer uses cffi at all)

pmrowla:
The flaky pypy tests should now be taken care of. The issue was a difference `thread.is_alive()` behavior in CPython vs pypy. The PR now blocks on explicit `threading.Event()` to signal that the writer thread is done rather than relying on `thread.is_alive()` which should resolve the pypy race condition.

pmrowla:
@jdavid we'd appreciate it if we could get a release for this pushed whenever you can, thanks!

jdavid:
yes, this week, just have to finish the ongoing PRs

jdavid:
The release is done

PR #1248, opened 2023-11-18
Submodule init, update, caching, reload, status

jorio:
This PR fleshes out submodules a bit.

- **Submodule.init & update**: Allow initing and updating Submodule instances when you already have them instead of having to go through `Repository.init_submodules` and `update_submodules` (which would do a redundant submodule lookup). Those two functions still exist in Repository, but they are now convenience wrappers around the new init/update functions in all the submodules in the repo.
- **Repository.submodule_cache_all & submodule_cache_clear**: Expose libgit2's submodule caching features. Speeds up submodule lookups which can be very slow in some repos.
- **Submodule.reload**: Reload cached info in a Submodule instance if you have reason to believe it has changed externally. Removes the need to do a new submodule lookup to get fresh information.
- **Repository.submodule_status**: Get the status of a submodule (is it uninitialized, does its workdir have any changes, did the head move, etc.).

---

Notes about API design: The PR brings in new enums (GIT_SUBMODULE_STATUS_...) that I encapsulated as enum classes (IntFlag, IntEnum) in `enums.py`.

This allows for friendlier development – function signatures can say they return a `SubmoduleStatus` instead of just an `int`, and calling `repr` on a `SubmoduleStatus` returns more descriptive text than just a number, such as:
`<SubmoduleStatus.IN_HEAD|IN_INDEX|IN_CONFIG|WD_UNINITIALIZED: 135>`

This isn't how pygit2 currently exposes enums to Python code – but if you think it'd improve pygit2, we could progressively convert the existing enums to IntFlag/IntEnum classes. (The raw GIT_* values would still be available for legacy code.)

If you'd rather not have `enums.py` for consistency with the existing API, let me know and I'll edit this PR accordingly.


jdavid:
Merged, thanks.

My only concern is the API consistency of the `Repository` class:

```
add_submodule
init_submodules
listall_submodules
lookup_submodule
update_submodules
submodule_status (NEW)
submodule_cache_all (NEW)
submodule_cache_clear (NEW)
```

I think it would be best to do like the remotes and have:

```
.submodules.add
.submodules.init
.submodules.__iter__
.submodules.__getitem__
.submodules.update
.submodules.status
.submodules.cache_all
.submodules.cache_clear
```

With the old function names kept for backwards compatibility, as deprecated.

What do you think? Do you mind to make a new PR in that direction?

jorio:
I agree, this looks better. I'll submit a new PR this week!

PR #1250, opened 2023-11-22
Group submodule functions under Repository.submodules

jorio:
Following up on https://github.com/libgit2/pygit2/pull/1248#issuecomment-1818816288, this PR implements a `Submodules` class that groups submodule functions in a Repository, much like `Branches`, `References` and `RemoteCollection`.

For backward compatibility, I turned the old functions into simple pass-throughs to the new class. Should we write unit tests for the deprecated versions?

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -65,6 +65,7 @@ class BaseRepository(_Repository):
         self.branches = Branches(self)
         self.references = References(self)
         self.remotes = RemoteCollection(self)
+        self.submodules = Submodules(self)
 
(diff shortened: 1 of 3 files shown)


jdavid:
Great @jorio 

Could you please:

- Remove `submodule_status`, `submodule_cache_all` and `submodule_cache_clear` ; these are new so no need to go through the deprecation process.
- In `add_submodule`, `init_submodules`, `lookup_submodule` and `update_submodules` : call `warnings.warn(...)` ; see `pygit2/repository.py` line 323
- Move the `Submodules` class to its own file `submodules.py`

Thanks!

PR #1251, opened 2023-11-24
Usability improvement proposal: Pythonic enums

jorio:
This PR introduces "pythonic" IntEnum and IntFlag classes that can supersede `GIT_...` integers in user code.

The main benefit of these enum classes is that they naturally guide the user toward the set of flags you can pass to a function, instead of having the user trawl through a list of GIT_ constants. Some examples:

```patch
-repo.checkout(someref, strategy=GIT_CHECKOUT_FORCE)
+repo.checkout(someref, strategy=CheckoutStrategy.FORCE)

-repo.apply(patch, GIT_APPLY_LOCATION_INDEX)
+repo.apply(patch, ApplyLocation.INDEX)

-diff = tree1.diff_to_tree(tree2, GIT_DIFF_SHOW_BINARY | GIT_DIFF_INCLUDE_UNMODIFIED)
+diff = tree1.diff_to_tree(tree2, DiffFlags.SHOW_BINARY | DiffFlags.INCLUDE_UNMODIFIED)
```

Another benefit is that the new enums are augmented with docstrings. I'm a heavy user of pygit2 and I tend to frequently have to dive into libgit2's source code to recall what some flags do – the new enums reduce this friction:

Screenshot_20231124_230949

Notes:
- GIT_ constants are still available for backward compatibility.
- The enums transparently demote to ints where needed, including passing values to C functions that expect an int.
- IntFlag objects can be constructed with arbitrary ints. So, those are fine for forward compatibility. For example, if libgit2 ever returns a GIT_STATUS that is, say, 1<<31, and we don't support it yet, `StatusFlags(1<<31)` will still work and it'll convert to a valid int.
- IntEnum objects, on the other hand, can NOT be constructed with arbitrary ints. Extra care must be given when converting an int that comes from libgit2 to an IntEnum. Luckily, such cases are rare – so far the only occurrence is in `Repository.state()` (where I just return a raw int if it can't be converted to a RepositoryState).
- Additionally, this PR exposes DiffFlags.IGNORE_BLANK_LINES (GIT_DIFF_IGNORE_BLANK_LINES) a bunch of new checkout strategy constants that were previously missing from pygit2.

If you think this PR worth merging into pygit2, I'm happy to discuss naming choices, etc.

----

**Next steps:** If this PR is accepted, I will submit another one later that makes some C functions return actual Python enums. This way, functions such as `repo.status_file(...)` will return a real StatusFlags instead of an int, which would make it easier to toy with pygit2 in an REPL.
```patch
 >>> r.status_file('pygit2/callbacks.py')   # this is implemented in pure C
- now returns: 256      # still an int, but you can already compare it to StatusFlags transparently
+ will return: <StatusFlags.WT_MODIFIED: 256>       # next PR (if this one is accepted)
```


jdavid:
This is great @jorio , we just need to get the names right because we won't be able to change them later.

My first reflex would be to do a literal translation from the libgit2 constants, as you have done in some cases, like:

    GIT_APPLY_LOCATION_WORKDIR -> ApplyLocation.WORKDIR

But for others this is not so obvious. For some you have chosen to add the suffix `Flags`, like:

    GIT_ATTR_CHECK_FILE_THEN_INDEX -> AttrCheckFlags.FILE_THEN_INDEX

An alternative would be to namespace the enums and flags, for example:

    from pygit2 import flags
    flags.AttrCheck

Instead of:

    from pygit2 import AttrCheckFlags
    AttrCheckFlags

I think this approach would have the following benefits:

- Help with name consistency
- Reduce the clutter in the root namespace `pygit2.*`

What do you think?

jdavid:
For reference, if I have not made errors, the proposed name changes in this PR could be summarized like...

IntEnum:

    GIT_APPLY_LOCATION_WORKDIR          ApplyLocation.WORKDIR
    GIT_CONFIG_LEVEL_PROGRAMDATA        ConfigLevel.PROGRAMDATA
    GIT_DELTA_UNMODIFIED                DeltaStatus.UNMODIFIED
    GIT_DESCRIBE_DEFAULT                DescribeStrategy.DEFAULT
    GIT_FETCH_PRUNE_UNSPECIFIED         FetchPruneMode.UNSPECIFIED
    GIT_FILTER_TO_WORKTREE              FilterMode.TO_WORKTREE
    GIT_OPT_GET_MWINDOW_SIZE            LibraryOption.GET_MWINDOW_SIZE
    GIT_OBJECT_ANY                      GitObjectType.ANY
    GIT_REFERENCES_ALL                  ReferenceFilter.ALL
    GIT_REPOSITORY_INIT_SHARED_UMASK    RepositoryInitMode.SHARED_UMASK
    GIT_REPOSITORY_STATE_NONE           RepositoryState.NONE
    GIT_RESET_SOFT                      ResetType.SOFT
    GIT_STASH_APPLY_PROGRESS_NONE       StashApplyProgress.NONE
    GIT_SUBMODULE_IGNORE_UNSPECIFIED    SubmoduleIgnore.UNSPECIFIED

IntFlag:

    GIT_ATTR_CHECK_FILE_THEN_INDEX      AttrCheckFlags.FILE_THEN_INDEX
    GIT_BLAME_NORMAL                    BlameFlags.NORMAL
    GIT_BLOB_FILTER_CHECK_FOR_BINARY    BlobFilterFlags.CHECK_FOR_BINARY
    GIT_BRANCH_LOCAL                    BranchType.LOCAL
    GIT_CHECKOUT_NOTIFY_NONE            CheckoutNotifyFlags.NONE
    GIT_CHECKOUT_NONE                   CheckoutStrategy.NONE
    GIT_CREDENTIAL_USERPASS_PLAINTEXT   CredentialType.USERPASS_PLAINTEXT
    GIT_DIFF_FLAG_BINARY                DeltaFlags.BINARY
    GIT_DIFF_FIND_BY_CONFIG             DiffFindFlags.FIND_BY_CONFIG
    GIT_DIFF_NORMAL                     DiffFlags.NORMAL
    GIT_DIFF_STATS_NONE                 DiffStatsFlags.NONE
    GIT_FEATURE_THREADS                 FeatureFlags.THREADS
    GIT_FILEMODE_UNREADABLE             FileMode.UNREADABLE
    GIT_FILTER_DEFAULT                  FilterFlags.DEFAULT
    GIT_MERGE_ANALYSIS_NONE             MergeAnalysis.NONE
    GIT_MERGE_PREFERENCE_NONE           MergePreference.NONE
    GIT_REF_INVALID                     ReferenceType.INVALID
    GIT_REPOSITORY_INIT_BARE            RepositoryInitFlags.BARE
    GIT_REPOSITORY_OPEN_NO_SEARCH       RepositoryOpenFlags.NO_SEARCH
    GIT_REVSPEC_SINGLE                  RevSpecFlags.SINGLE
    GIT_STATUS_CURRENT                  StatusFlags.CURRENT
    GIT_SUBMODULE_STATUS_IN_HEAD        SubmoduleStatus.NONE
    GIT_SORT_NONE                       WalkerSortFlags.IN_HEAD

Let's agree on the names first.

jdavid:
Hi @jorio 

I agree with your points (all in `pygit2.enums` and the name must be unique).
Here a proposal, it's a synthesis of your two proposals:

```
ApplyLocation
AttrCheck
BlameFlag
BlobFilter
BranchType
CheckoutStrategy
CheckoutNotify
ConfigLevel
CredentialType
DeltaStatus
DescribeStrategy
DiffFind
DiffFlag
DiffOption
DiffStatsFormat
Feature
FetchPrune
FileMode
FilterFlag
FilterMode
MergeAnalysis
MergePreference
ObjectType
Option
ReferenceType
ReferenceFilter
RepositoryInitFlag
RepositoryInitMode
RepositoryOpenFlag
RepositoryState
ResetMode
RevSpecFlag
SortMode *
StashApplyProgress
FileStatus 
SubmoduleIgnore
SubmoduleStatus
```

Only change of mine is that I've changed `WalkerSort` to `SortMode`.

So the general rule would be:

- names based on the typedef,
- shortened when that does not make the name ambiguous,
- when the name is ambiguous, added a suffix: either `Flag`, `Type` or `Mode`

With few exceptions: `FileStatus` because it makes sense, `Option` instead of `Opt`.

If the list of names above looks good, then you can go ahead with the PR. Or tell me if you still want to make changes.
Thanks

jorio:
Hi @jdavid, this looks good to me! I'll amend the PR with the new names soon-ish. Thanks again.

jorio:
Hi @jdavid, I rebased the PR on master with the name changes that you suggested.

I also removed the individual enum classes from the root namespace per your suggestion.

Furthermore, I moved the GIT_... constants (for backwards compatibility) into legacyenums.py to make \_\_init\_\_.py a bit lighter. I don't mind undoing this if you prefer having fewer files.

Let me know if there's anything that still needs tweaking. Thank you!

jdavid:
Thanks @jorio !

PR #1257, opened 2023-12-08
Fix type hints not found by Mypy

### LINKED ISSUE #1256: Type hints are not found by Mypy
Hi!

Running `mypy` (1.7.1) on file importing `pygit2` (1.13.3) causes the following error:

```
$ mypy file.py
file.py:1: error: Skipping analyzing "pygit2": module is installed, but missing library stubs or py.typed marker  [import-untyped]
file.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 3 source files)
```

It's surprising because `pygit2` [does have a stubs file](https://github.com/libgit2/pygit2/blob/37b0ce1e00da1ad3ba944221761431c72dfa2b56/pygit2/_pygit2.pyi).

I think this is because `py.typed` must be added and installed.

Delgan:
This should fix #1256.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -146,7 +146,7 @@ setup(
     maintainer_email='jdavid.ibp@gmail.com',
     long_description=long_description,
     packages=['pygit2'],
-    package_data={'pygit2': ['decl/*.h', '*.pyi']},
+    package_data={'pygit2': ['decl/*.h', '*.pyi', 'py.typed']},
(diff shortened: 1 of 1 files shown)


Delgan:
Closing as I now realize the type hints are intended for the internal C API, not the end-user. Sorry.

PR #1268, opened 2024-01-11
Fix git-cherry-pick.rst example.

herrerog:
Using the parent of the cherry-picked commit as ancestor is not correct sometimes. That looks to be an issue because `base` was not used at all in the example whereas it should be used as ancestor.

### DIFF
--- a/docs/recipes/git-cherry-pick.rst
+++ b/docs/recipes/git-cherry-pick.rst
@@ -51,10 +51,9 @@ example `three-argument rebases`_.
     cherry = repo.revparse_single('9e044d03c')
     basket = repo.branches.get('basket')
 
-    base      = repo.merge_base(cherry.id, basket.target)
-    base_tree = cherry.parents[0].tree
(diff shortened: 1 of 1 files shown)


herrerog:
Sorry I got that one wrong actually @jdavid 
We should use `cherry.parents[0].tree` as ancestor otherwise the resulting merge may be wrong.
Here what needs to be used:
```
repo = pygit2.Repository('/path/to/repo')

cherry = repo.revparse_single('9e044d03c')
basket = repo.branches.get('basket')

base_tree = cherry.parents[0].tree

index = repo.merge_trees(base_tree, basket, cherry)
tree_id = index.write_tree(repo)

author    = cherry.author
committer = pygit2.Signature('Archimedes', 'archy@jpl-classics.org')

repo.create_commit(basket.name, author, committer, cherry.message,
                   tree_id, [basket.target])
del None # outdated, prevent from accidentally using it
```

Let me know if I should send another pull request.

jdavid:
@herrerog Yes please

PR #1271, opened 2024-01-17
Make merge API consistent with new enums

jorio:
Merge functions now use new enums: MergeFavor, MergeFlag, MergeFileFlag.

These functions used to take a dict of strings to boolean flags (e.g. `flags={'no_recursive': True}`). Backward compatibility with this system was kept, and there are unit tests for that too.

As far as I'm aware, this system was only used in the merge functions and not anywhere else in pygit2. So, this PR aims to make these functions more consistent with the rest of the library, now that we use enums elsewhere.

### DIFF
--- a/docs/merge.rst
+++ b/docs/merge.rst
@@ -19,10 +19,12 @@ branch reference in the case of a fastforward.
 
 Example::
 
+    >>> from pygit2.enums import MergeFavor, MergeFlag
     >>> other_branch_tip = '5ebeeebb320790caf276b9fc8b24546d63316533'
     >>> repo.merge(other_branch_tip)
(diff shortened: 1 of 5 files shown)


jdavid:
@jorio Shouldn't this:

    def merge_commits(
            [...]
            file_flags = MergeFavor.NORMAL,

https://github.com/libgit2/pygit2/blob/master/pygit2/repository.py#L700

be `MergeFileFlag.DEFAULT` ?

jorio:
@jdavid My bad, you're right (it resolves to 0 either way so the tests didn't fail). PR https://github.com/libgit2/pygit2/pull/1272 fixes this.

PR #1280, opened 2024-03-04
Format the repository and enforce code style in CI

xen0n:
No functional change; this is in preparation of more refactoring for better typing within the library.

Context: https://github.com/python/typeshed/pull/11374 -- it turns out it's probably impossible to make pyright happy with the library with stubs alone.


jdavid:
Great, thanks!
Most changes are because of the use of single vs double quotes. But I think this is configurable in ruff, with `quote-style = "single"` from what I've read. Can you change this to make the PR smaller?

xen0n:
> Great, thanks! Most changes are because of the use of single vs double quotes. But I think this is configurable in ruff, with `quote-style = "single"` from what I've read. Can you change this to make the PR smaller?

Thank you for the suggestion! I've checked how `quote-style = "single"` would fare though, it's `+1996 -1136` compared with `+2929 -2069` in commit 70e74755b73a -- it seems the quote style is like 50-50 split while the other changes (mostly indentation and line wrapping changes) are equally substantial!

Given double quotes are the default `black` style, would keeping that style be acceptable in exchange for not having to customize anything?

jdavid:
Hi, I still prefer single quotes, the smaller change the better, this will not obfuscate git blame too much.

xen0n:
> Hi, I still prefer single quotes, the smaller change the better, this will not obfuscate git blame too much.

Okay, will do it!

PR #1298, opened 2024-06-02
Add Repository.hashfile()

jorio:
This exposes libgit2's [`git_repository_hashfile()`](https://libgit2.org/libgit2/#HEAD/group/repository/git_repository_hashfile).

`pygit2.hashfile()` already exists, but this lets you hash files using the repository's filters.

### DIFF
--- a/pygit2/decl/repository.h
+++ b/pygit2/decl/repository.h
@@ -83,6 +83,7 @@ int git_repository_set_head_detached(
 	git_repository* repo,
 	const git_oid* commitish);
 
+int git_repository_hashfile(git_oid *out, git_repository *repo, const char *path, git_object_t type, const char *as_path);
 int git_repository_ident(const char **name, const char **email, const git_repository *repo);
 int git_repository_set_ident(git_repository *repo, const char *name, const char *email);
 int git_repository_index(git_index **out, git_repository *repo);
--- a/pygit2/decl/types.h
+++ b/pygit2/decl/types.h
@@ -57,3 +57,14 @@ typedef enum {
 	GIT_SUBMODULE_IGNORE_DIRTY     = 3,
 	GIT_SUBMODULE_IGNORE_ALL       = 4,
 } git_submodule_ignore_t;
+
+typedef enum {
+	GIT_OBJECT_ANY       = ...,
+	GIT_OBJECT_INVALID   = ...,
+	GIT_OBJECT_COMMIT    = ...,
+	GIT_OBJECT_TREE      = ...,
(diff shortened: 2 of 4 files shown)


jdavid:
It looks good, but the tests don't pass on Windows:
https://ci.appveyor.com/project/jdavid/pygit2/builds/49936765/job/72q19bou1q95mxy9

jorio:
Hmm, this looks like a libgit2 issue on Windows.

`str(Path(...))` returns a "native" path that uses `\` as the directory separator. But `git_repository_hashfile` doesn't recognize this path as being part of the workdir – which is required for the test to pass – unless you replace the backslashes with forward slashes.

In other words, editing my test as follows will make it pass on Windows:
```python
    # Treat absolute path with filters
    absolute_path = str(Path(testrepo.workdir, 'hellocrlf.txt'))
    absolute_path = absolute_path.replace('\\', '/')  # <--------------- yikes!
    h = testrepo.hashfile(absolute_path)
    assert h == original_hash
```

But in this case, I would argue that libgit2 should canonicalize the input path to its preferred form internally.

jdavid:
Thanks for the excellent report in libgit2

I think we can just skip the test on Windows, with a reference to the libgit2 issue, and then merge the PR.

jorio:
If we want to keep the test on Windows, pygit2 could sanitize the path before passing it to libgit2 on Windows, and perhaps issue a RuntimeWarning with a link to the issue if the path contains any backslashes.

That would make hashfile work in the "least surprising way" on Windows. I can add this to the PR soon if we think that's a good compromise. What do you think?

jdavid:
The PR is straightforward and the error is clearly in the libgit2 side.
What you propose is in the best interest of users, yet I'm not sure we should workaround a libgit 2 bug.
Do as you see fit, I leave it to your good judgement.

jdavid:
By the way you have been the largest contributor to pygit2 since you started few years ago, about half the contributions no less. Also your contributions have been of an outstanding quality, making my reviews easy.
So I would like to give you a big thanks!

jorio:
Glad to be of service! And thank you for steering this project in the right direction :)

I have a vested interest in improving pygit2 because I’ve been working on a GUI that hits almost all of pygit2’s API surface area. I’m planning to make the first public release soon-ish. Here’s what it looks like:

Untitled2

Anyway, regarding this PR - let’s leave the implementation as it is, but the unit test could demonstrate how to make it work on Windows (call Path.as_posix() before passing the path to hashfile). I rebased my branch and fixed the test, hopefully it passes on Windows now.

Issue #945, opened 2019-09-26
'certificate' parameter in RemoteCallbacks doesn't work


nu11ptr:
Looking at the code, I think the 'certificate' __init__ param is supposed to be named 'certificate_check' so that if it is set, it overrides the class function (I believe this is how 'credentials' is working). As it stands, I don't see anywhere in the class where 'certficate' is referenced (fetch_opts it only references 'certificate_check') and some tests to self-signed servers didn't work until I created my own subclass with my own version of 'certificate_check'.

Issue #1262, opened 2023-12-30
`certificate_check` callback not used in `remote.connect` or `remote.ls_remotes`


kevinvalk:
As far as I can see the remote_callbacks struct is not properly initialized with the `certificate_check` callback (https://libgit2.org/libgit2/#v0.23.2/type/git_remote_callbacks).

https://github.com/libgit2/pygit2/blob/611a3b3462a1acb6bc3ca0d9bd3eff7bf6154131/pygit2/callbacks.py#L384-L387

This makes it impossible to "bypass" certificate checks for example when testing remotes. So I think simply adding to the callback options is enough, but do not quote me on that!
```python
cdata.certificate_check = C._certificate_cb
```

This also may be the problem for https://github.com/libgit2/pygit2/issues/945

P.S. Its point about `certificate_check` vs `certificate` still stands so that could also be sorted out.

kevinvalk:
Super ugly monkey patch to get `certificate_check` working in version 1.13.3 of pygit2
```python
class MyCallbacks(pygit2.RemoteCallbacks):
    def certificate_check(testrepo, certificate, valid, host):
        logging.error("Yes I bypass stuff")
        return True

    def __setattr__(self, name: str, value: Any) -> None:
        if name == "remote_callbacks":
            from pygit2.ffi import C

            value.certificate_check = C._certificate_cb
        return super().__setattr__(name, value)
```

PR #1321, opened 2024-10-10
Fix certificate check callback

sathieu:
Fixes: #945 #1262


sathieu:
Note: this looks like a breaking change, but from a non-working feature, I think this is acceptable.

PR #1326, opened 2024-11-20
feat: add certificate_check callback to ls_remotes

kevinvalk:
Fixes: https://github.com/libgit2/pygit2/issues/1262

I also added a test to ensure the callback is called on `ls_remotes`.


kevinvalk:
@jdavid, formatted using ruff now, should (hopefully) pass CI.

PR #1327, opened 2024-11-24
Typo in repository.c

johannkm:


jdavid:
Thanks @johannkm but there is still a typo, it should be "do not" instead of "no not"

johannkm:
Oops, haha

PR #1337, opened 2025-01-04
Pypy wheel macos - build on macOS 10.15

Spitfire1900:
Per cibuildwheel's documentation PyPy 3.9+ support is only available on macOS 10.15.

I don't know if this breaks compatibility with runtimes on 10.14, 10.15 though :slightly_frowning_face:

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - master
+    - pypy-wheel-macos
     tags:
     - 'v*'
 
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -33,7 +33,7 @@ jobs:
           platforms: all
 
       - name: Install cibuildwheel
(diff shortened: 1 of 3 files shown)


Spitfire1900:
The following error is thrown when attempting to build the wheel, I'm not familiar enough to know if this is because of in issue in PyGit2's build configuration or an issue in cibuildwheel itself.

```
NotImplementedError: This function does not support separate values per-architecture: {'x86_64': [('@rpath/libcrypto.3.dylib', '3.0.0', '3.0.0'), ('/usr/lib/libSystem.B.dylib', '1.0.0', '1351.0.0')], 'arm64': [('@rpath/libcrypto.3.dylib', '3.0.0', '3.0.0'), ('/usr/lib/libz.1.dylib', '1.0.0', '1.2.12'), ('/usr/lib/libSystem.B.dylib', '1.0.0', '1351.0.0')]}
```

Spitfire1900:
Opened https://github.com/pypa/cibuildwheel/issues/2121 on this

PR #1338, opened 2025-01-05
Free threaded wheel

Spitfire1900:
Add wheels for free threaded Python.

Currently blocked on cffi adding free threaded support.


webknjaz:
> Currently blocked on cffi adding free threaded support.

v2.0.0b1 was released with this on Jul 29. So while it's still a pre-release, it enables you to start testing against it.

PR #1348, opened 2025-02-07
Allow passing a Reference to Repository.merge()

jorio:
This PR allows passing more types to Repository.merge. Previously, you could only pass a commit ID as an Oid or an str.

In addition to these two, you can now also pass a real Commit, or a Reference.

Passing in a Reference will enrich the merge with additional information. For example, Repository.message will specify the name of the branch being merged, instead of just stating a commit ID:

```python
branch = repo.branches.local['other']
repo.merge(branch)
print(repo.message)  # "Merging branch 'other'"
```

If this PR is accepted, I believe passing an str should now be deprecated. The str is supposed to represent a commit ID, but users might intuit that the function also accepts ref names as an str, when in fact they should pass a real Reference object.

### DIFF
--- a/pygit2/decl/commit.h
+++ b/pygit2/decl/commit.h
@@ -13,4 +13,9 @@ int git_annotated_commit_lookup(
 	git_repository *repo,
 	const git_oid *id);
 
+int git_annotated_commit_from_ref(
+	git_annotated_commit **out,
+	git_repository *repo,
+	const struct git_reference *ref);
+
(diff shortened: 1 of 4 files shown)


jdavid:
> If this PR is accepted, I believe passing an str should now be deprecated. The str is supposed to represent a commit ID, but users might intuit that the function also accepts ref names as an str, when in fact they should pass a real Reference object.

Agree, there should be a deprecation warning, could you make a PR?

PR #1332, opened 2024-12-22
random fixes

DinhHuy2010:
adding more type checking into the library

This will add typing information since pygit2 stubs is obsolete at https://github.com/python/typeshed/pull/12785
Also others list at the tasks:

# Tasks/TODO

 - [x] Add typing to pygit2/_pygit2.pyi
 - [x] Add pygit2/_libgit2.pyi
 - [ ] Add typing to pygit2/settings.py
 - [ ] Add typing to pygit2/__init__.py
 - [x] Add typing to pygit2/errors.py
 - [ ] Add typing to pygit2/remotes.py
 - [ ] Add typing to pygit2/submodules.py
 - [x] Add typing to pygit2/packbuilder.py
 - [ ] Add typing to pygit2/refspec.py
 - [ ] Add typing to pygit2/callbacks.py
 - [ ] Add typing to pygit2/repository.py
 - [x] Add typing to pygit2/branches.py
 - [x] Add typing to pygit2/utils.py
 - [x] Add typing to pygit2/credentials.py
 - [x] Add typing to pygit2/enums.py
 - [x] Add typing to pygit2/_run.py
 - [ ] Add typing to pygit2/references.py
 - [x] Add typing to pygit2/config.py
 - [x] Add typing to pygit2/_build.py
 - [ ] Add typing to pygit2/legacyenums.py
 - [x] Add typing to pygit2/ffi.py
 - [x] Add typing to pygit2/blame.py
 - [x] Add typing to pygit2/blob.py
 - [ ] Add typing to pygit2/index.py
 - [x] Add more ruff configuration
 - [x] Lint and format using ruff
 - [x] Add pyright configuration for type checking
 - [ ] Add more GitHub Actions workflows
 - [ ] Maybe update docs

PR #1350, opened 2025-02-17
typecheck-part-1

DinhHuy2010:
Split from https://github.com/libgit2/pygit2/pull/1332


pyproject.toml
```diff
@@ -25,15 +25,53 @@ environment = {LIBGIT2_VERSION="1.9.0", LIBSSH2_VERSION="1.11.1", OPENSSL_VERSIO
...
+    "I",
+    "B",
+    "C4",
+    "ARG",
+    "SIM",
+    "PTH",
+    "PL",
+    "TID",
+]
+lint.ignore = [
```
Tsafaras, line 44: How come you're not using this instead, for `select` and `ignore`?
```toml
[tool.ruff.lint]
```
As seen (and suggested one might say) in [ruff's docs](https://docs.astral.sh/ruff/configuration/#configuring-ruff).

pyproject.toml
```diff
@@ -25,15 +25,53 @@ environment = {LIBGIT2_VERSION="1.9.0", LIBSSH2_VERSION="1.11.1", OPENSSL_VERSIO
 repair-wheel-command = "DYLD_LIBRARY_PATH=/Users/runner/work/pygit2/pygit2/ci/lib delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}"
 
 [tool.ruff]
+target-version = "py310" # oldest supported Python version
+fix = true
```
Tsafaras, line 29: Shouldn't `fix`es be opt-in, instead of opt-out? 🤔 Would you mind elaborating on your reasoning?

pyproject.toml
```diff
@@ -25,15 +25,53 @@ environment = {LIBGIT2_VERSION="1.9.0", LIBSSH2_VERSION="1.11.1", OPENSSL_VERSIO
...
+    "W293",    # Blank line contains whitespace
+    "PLR0912", # Too many branches
+    "PLR2004", # Magic values
+    "PLR0915", # Too many statements
+    "PLW0603", # Global statement
+    "PLR0913", # Too many arguments
+    "B010",    # setattr
+    "F401",    # unused imports
+    "ARG002",  # unused arguments
+    "SIM105",  # try-except-pass
```
Tsafaras, line 56: Are you ignoring them because there are many fixes to be made?

If so, the suggested approach is not to ignore them at the configuration level, but run ruff with `--add-noqa`, as seen [here](https://docs.astral.sh/ruff/tutorial/#adding-rules).

> When enabling a new rule on an existing codebase, you may want to ignore all _existing_ violations of that rule and instead focus on enforcing it going forward.

pyproject.toml
```diff
@@ -25,15 +25,53 @@ environment = {LIBGIT2_VERSION="1.9.0", LIBSSH2_VERSION="1.11.1", OPENSSL_VERSIO
...
+]
 
 [tool.ruff.format]
 quote-style = "single"
 
+[tool.pyright]
```
Tsafaras, line 62: You are adding a `pyright` configuration but `pyright` is not in the repo in any capacity.

DinhHuy2010:
hi, sorry for waiting
yes I probably will make a new PR'

and yes @Tsafaras, I forgot to change to new style (and also did not know about the `--add-noqa` argument)

Alright, Let me fix the changes.

Thanks.

pyproject.toml
DinhHuy2010: oops, that should be opt-in, sorry.

pyproject.toml
DinhHuy2010: Ah, never know that kind of thing, thanks!

pyproject.toml
DinhHuy2010: will add pyright soon in another PR

jdavid:
Closing as superseded by 84310f3c and later commits.
Open to new and focused PRs.

PR #1356, opened 2025-03-18
doc: warn against using init_repository() to open a repository.

adrien-n:
It turns out this works in some situations but it should be avoided.

### DIFF
--- a/pygit2/__init__.py
+++ b/pygit2/__init__.py
@@ -106,6 +106,10 @@ def init_repository(
     The *workdir_path*, *description*, *template_path*, *initial_head* and
     *origin_url* are all strings.
 
+    If a repository already exists at *path*, it may be opened successfully but
+    you must not rely on that behavior and should use the Repository
+    constructor directly instead.
+
     See libgit2's documentation on git_repository_init_ext for further details.
     """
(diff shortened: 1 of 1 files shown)


adrien-n:
I forgot to mention I'm also using that PR as a question. Maybe I got that wrong, I'm not sure, but worktrees don't work with init_repository(), or at least not with every flag or combination of flags.

jdavid:
I don't know, we just call `git_repository_init_ext`; but the comment makes sense to me, and it matches my interpretation of `git_repository_init_ext` docs:

> Create a new Git repository in the given folder with extended controls.
> 
> This will initialize a new git repository [...]

https://libgit2.org/docs/reference/main/repository/git_repository_init_ext.html

adrien-n:
I noticed the issue with some software using pygit2 and started investigating. I think this also applies to libgit2 directly but I haven't fully followed all the code paths and I haven't found the time to reproduce the issue with it only (which is more involved than comparing Repository.('.') and init_repository('.') in a REPL).

I'll also create a similar PR for libgit2's documentation. I think it's worth having the warning in both places since it appears to be an actual real-world issue which may go un-noticed simply because worktree usage is less widespread.

PR #1360, opened 2025-04-09
repository - diff blobs: consider the values of context_lines and interhunk_lines

eantoranz:


src/blob.c
```diff
@@ -70,11 +70,12 @@ Blob_diff(Blob *self, PyObject *args, PyObject *kwds)
...
     int err;
-    char *keywords[] = {"blob", "flag", "old_as_path", "new_as_path", NULL};
+    char *keywords[] = {"blob", "flag", "old_as_path", "new_as_path", "context_lines", "interhunk_lines", NULL};
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O!Iss", keywords,
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|O!IssHH", keywords,
                                      &BlobType, &other, &opts.flags,
-                                     &old_as_path, &new_as_path))
+                                     &old_as_path, &new_as_path,
+                                     &opts.context_lines, &opts.interhunk_lines))
```
jdavid, line 78: This adds 2 new keyword arguments to Blob.diff but the documentation string is not updated, and please add unit tests for the new arguments.

jdavid (changes requested):

src/blob.c
eantoranz: Just pushed an amended commit with the required adjustments.

test/test_diff.py
```diff
@@ -107,6 +107,84 @@
  delete mode 100644 c/d
 """
 
+TEXT_BLOB1 = """Common header of the file
```
eantoranz, line 110: I thought it was much simpler to create blobs out of thin air than... _hunt down_ for usable examples in the repos.

PR #1366, opened 2025-04-15
ci: change triggers to prevent dual runs on prs

clintonsteiner:
* runs on pushes to master or prs to master
* will prevent action from running for the push in pr and pull request into master

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -4,7 +4,11 @@ name: Codespell
 
 on:
   pull_request:
+    branches:
+    - master
   push:
+    branches:
+    - master
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


webknjaz:
Personally, I'd recommend disallowing topic branches upstream via branch rulesets and forcing them into forks. This would solve said problem while still allowing branches in forks to be tested on pushes.

jdavid:
Tests and checks are fast, about 2 minutes. The slow actions (wheels, s390x tests) are only run on push to the master branch. I don't see this to be much of a problem, do I miss something?

webknjaz:
@jdavid this also only affects PRs from upstream, not forks. But in general, it's a good organizational decision to separate topic branches to they don't collect dust upstream.

PR #1370, opened 2025-05-03
Typing fixes - remove deleted symbols from stub

Andrej730:
See commits descriptions for the details.

@jdavid can you please take a look?

### DIFF
--- a/pygit2/_pygit2.pyi
+++ b/pygit2/_pygit2.pyi
@@ -36,10 +36,8 @@ LIBGIT2_VER_REVISION: int
 class Object:
     _pointer: bytes
     filemode: FileMode
-    hex: str
     id: Oid
     name: str | None
-    oid: Oid
     raw_name: bytes | None
     short_id: str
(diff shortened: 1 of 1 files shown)


Andrej730:
Fixed linter issue.

pygit2/__init__.py
```diff
@@ -61,6 +61,8 @@
 from .utils import to_bytes, to_str
 
 
+__all__ = ('Repository',)
```
jdavid, line 64: There is more in pygit2 than `Repository`. Please drop this change from this PR and I will merge it, and then maybe open a new PR to define `__all__`.

jdavid (changes requested):

pygit2/__init__.py
Andrej730: Sure. 
I'm not that familar with pygit2 - what else should be added to `__all__`? The only issue I've met was related with `Repository`. Though I think submitting this change I've missed that `__all__` may also affect the users that rely on `from pygit2 import *`.

If it's not clear at the moment what should go to `__all__`, then maybe we can hide it under `TYPE_CHECKING` to avoid any runtime implications?

```python
if typing.TYPE_CHECKING:
    __all__ = ("Repository",)
```

Andrej730: Created a separate PR https://github.com/libgit2/pygit2/pull/1371

Issue #1322, opened 2024-10-15
Missing types in v1.16


sathieu:
The addition of https://github.com/libgit2/pygit2/pull/1310 [made my CI fail](https://gitlab.com/gitlabracadabra/gitlabracadabra/-/jobs/8085594991) with the following:

```
gitlabracadabra/gitlab/pygit2.py:84: error: Call to untyped function "UserPass" in typed context  [no-untyped-call]
gitlabracadabra/gitlab/pygit2.py:86: error: Call to untyped function "UserPass" in typed context  [no-untyped-call]
gitlabracadabra/gitlab/pygit2.py:88: error: Call to untyped function "UserPass" in typed context  [no-untyped-call]
gitlabracadabra/gitlab/pygit2.py:90: error: Call to untyped function "UserPass" in typed context  [no-untyped-call]
gitlabracadabra/gitlab/pygit2.py:100: error: Call to untyped function "RemoteCallbacks" in typed context  [no-untyped-call]
gitlabracadabra/gitlab/pygit2.py:101: error: Cannot assign to a method  [method-assign]
gitlabracadabra/gitlab/pygit2.py:101: error: Incompatible types in assignment (expression has type "Callable[[str, str | None, CredentialType], UserPass]", variable has type "Callable[[str, str | None, CredentialType], Any]")  [assignment]
gitlabracadabra/gitlab/pygit2.py:103: error: Cannot assign to a method  [method-assign]
gitlabracadabra/objects/object.py:71: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
gitlabracadabra/mixins/mirrors.py:57: error: "Repository" has no attribute "remotes"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:59: error: "Repository" has no attribute "remotes"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:64: error: "Repository" has no attribute "remotes"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:65: error: "Repository" has no attribute "remotes"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:66: error: "Repository" has no attribute "remotes"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:67: error: "Repository" has no attribute "config"  [attr-defined]
gitlabracadabra/mixins/mirrors.py:82: error: "Repository" has no attribute "remotes"  [attr-defined]
[...]

PR #1371, opened 2025-05-03
export symbols using `__all__` to avoid typing errors

Andrej730:
Mentioned in #1322

Explanation on how it works - https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

Example failing snippet:
```python
import pygit2

# "Repository" is not exported from module "pygit2" PylancereportPrivateImportUsage
def test(repo: pygit2.Repository):
    ...
```

Wanted to solve that particular error with missing `Repository` symbol and I'm not very familar with pygit2 repository - so please advice if any symbols should be removed or added. 
Currently I've included all symbols that are imported in `__init__`, except `ffi`, `C` (looked like something private), to make it fully compatible with current `from pygit2 import *` behaviour (definingi `__all__` is limiting wild card import).


jdavid:
```python
import sys
before = set(dir())
from pygit2 import *
after = set(dir())
imported_symbols = after - before
print(len(imported_symbols))
```

This prints 412, while in this PR `__all__` has 127 symbols.
For backwards compatibility you can just put all of them, sorted alphabetically, then I will go through and remove those deemed internal or deprecated.

Also put `__all__` at the very end of the file.

PR #1382, opened 2025-06-27
index: map git_index_conflict_add

eantoranz:
map libgit2's `git_index_conflict_add()` as `Index.add_conflict()`.


eantoranz:
Interesting that it failed for macos (and that the feedback already tips that it might be a problem in libgit2).

pygit2/index.py
```diff
@@ -223,6 +223,41 @@ def add(self, path_or_entry):
...
+        if ancestor is not None:
+            centry_ancestor, _ = ancestor._to_c()
+        if ours is not None:
+            centry_ours, _ = ours._to_c()
+        if theirs is not None:
+            centry_theirs, _ = theirs._to_c()
+        err = C.git_index_conflict_add(
+            self._index, centry_ancestor, centry_ours, centry_theirs
+        )
+
```
eantoranz, line 258: Is there something to do with the `centries` after using them in `git_index_conflict_add()`?

jdavid:
Reading cffi's documentation, centries will be garbage collected, so there is nothing to do with it.

jdavid:
> Interesting that it failed for macos (and that the feedback already tips that it might be a problem in libgit2).

What do you mean? I don't see a clue about libgit2 here:

```python
        conflict = index.conflicts['conflict.txt']
>       assert conflict[0].id == ancestor_blob_id
               ^^^^^^^^^^^^^^
E       AttributeError: 'NoneType' object has no attribute 'id'
test/test_index.py:331: AttributeError
```

eantoranz:
> > Interesting that it failed for macos (and that the feedback already tips that it might be a problem in libgit2).
> 
> What do you mean? I don't see a clue about libgit2 here:

It is not me who tipped about `libgit2` being the culprint. It's [feedback from the failed test][2], actually:
```
=========================== short test summary info ============================
SKIPPED [1] test/test_credentials.py:191: Requires proxy at port 8888
SKIPPED [1] test/test_repository.py:741: Requires proxy at port 8888
XFAIL test/test_nonunicode.py::test_nonunicode_branchname - probably a bug in libgit2 for non-linux platforms <==== HERE
FAILED test/test_index.py::test_add_conflict - AttributeError: 'NoneType' object has no attribute 'id'
====== 1 failed, 524 passed, 2 skipped, 1 xfailed, 10 warnings in 29.74s =======
Error: Process completed with exit code 1.
```

And given that the other platforms pass having the exact same code on pygit2 (and not much room for behavior differences there), I can only assume that the call to libgit2's `git_index_conflict_add` inside `Index.add_conflict` [here][1] is providing a `NULL` value where it is actually not expected. So... I think the description from the tests feedback pointing at `libgit2` makes sense in this context. Am I missing something?

Just in case, what should be the right way to handle this? Should I go into libgit2, find the root cause of the test failure there, fix it and create a PR? It could get tricky given that I do not have a mac at hand so all I could do is just read code and try to make some (theoretical) sense as to why it might be failing.

[1]: https://github.com/libgit2/pygit2/pull/1382/files#diff-21c27b991511d3d0e3a14b4dfb56017689de2fa114c3edf11c577bd1ce744b72R326
[2]: https://github.com/libgit2/pygit2/actions/runs/15934930694/job/45422717425?pr=1382

jdavid:
Ah didn't scroll to the bottom. That test message is actually not correct, it was later identified to be an issue related to the filesystem, not the os, see https://github.com/libgit2/pygit2/pull/1329#issuecomment-2599690561

In our case now, it's not clear to me whether the error is in pygit2 or libgit2. True, pygit2's code is straightforward, but so looks libgit2 code, and it has an extensive test suite for this feature.

If it's an issue specific to macos, I'm fine adding an xfail, but maybe there is for example a memory bug in pygit2 that happens to trigger in the macos test suite, but that is present in Linux as well. I don't have a macos either, that makes things more difficult.

eantoranz:
Just rebased... let's see if that helps with the MacOS test.

PR #1395, opened 2025-08-04
Instrument `git_submodule_set_url`.

nrybowski:
This commit adds a setter for the  `url` property of the `Submodule` class. The new setter leverages `git_submodule_set_url` from libgit2.


jdavid:
Looks good. Could you add a unit test?

PR #1397, opened 2025-08-05
Improve usability of `Remote.ls_remotes`'s result

jorio:
`Remote.ls_remotes` used to return a dict with str keys, which is inconsistent with how pygit2 generally returns structured results. This PR introduces `RemoteHead`, a class with fields.

Backwards compatibility with the old-style dict was kept via `RemoteHead.__getitem__` which emits a `DeprecationWarning`.

Also, when the `local` flag is off, `RemoteHead.loid` now forwards oid zero (`0000000`) as returned by libgit2 instead of making it `None`.

--- 

Tangent: I think `ls_remotes` is a bit of a confusing name. Perhaps something like `list_heads` or `advertised_heads` would be more intuitive, but this could be for another PR. What do you think?

### DIFF
--- a/pygit2/remotes.py
+++ b/pygit2/remotes.py
@@ -25,7 +25,8 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Generator, Iterator, Literal, TypedDict
+import warnings
+from typing import TYPE_CHECKING, Any, Generator, Iterator, Literal
 
 # Import from pygit2
 from pygit2 import RemoteCallbacks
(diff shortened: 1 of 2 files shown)


jdavid:
`Remote.list_heads()` looks good to me, since it wraps `git_remote_ls` which returns a list of `git_remote_head`

jdavid:
@jorio 
What do you think about this?
- Add a new method `Remote.list_heads(...)` that returns a list of `RemoteHead` objects, these won't implement the mapping interface
- Keep the old `Remote.ls_remotes(...)` returning a list of dicts, as deprecated

jorio:
@jdavid Sorry for the late reply. This sounds sensible, please see https://github.com/libgit2/pygit2/pull/1410

PR #1404, opened 2025-08-08
Fix mypy strict equality

kbaikov:


kbaikov:
Ping @jdavid  any comments/review to this PR?

Issue #1046, opened 2020-11-09
Wheel release issues


SpaceManiac:
I use pygit2 in a Python tool which is distributed to non-programmers. Every new Python release, I get support requests from people who installed the latest Python version and the ever-so-slightly old version of pygit2 in the tool's `requirements.txt`. This has happened with 3.7, 3.8, and now 3.9.

Unfortunately the error messages for this situation are beyond useless, instructing users to install MSVC (a huge waste of time) and then to install libgit2 development headers (difficult to figure out how to do this).

Would it be feasible to either

- back-publish 3.9 wheels for old pygit2 versions; or
- use [ctypes](https://docs.python.org/3/library/ctypes.html?highlight=ctypes#module-ctypes) to allow for Python-version-independent binary wheels?

Regarding the latter, if you think it's feasible, I may be able to contribute some of the actual work.

segevfiner:
cffi might be capable of using the limited API/stable ABI. This means it might be possible to publish a wheel that works on multiple Python 3 versions including newly released ones.

encukou:
Python 3.10 will be released soon-ish.

The ABI for Python 3.10 is frozen since [the first release candidate](https://www.python.org/downloads/release/python-3100rc1/) was released. So it should be possible to build wheels for 3.10 – it is as safe to use 3.10.0rc1 wheels for 3.10.0 final as it is to use 3.9.1 wheels for 3.9.2.

PR #1409, opened 2025-08-11
Implement (almost) all global options

ambv:
This pull request is split into two commits, doing those things:

### Commit 1: moves option handling from the Python C API to CFFI
This simplifies future edits and removes some static typing weirdness around the `_pygit2.pyi` file needing Python-level enums to express the C-level functional API.

Mostly, this moves pygit2 in the direction of being more CFFI, which will eventually lead to beautiful things like not needing to push new wheels every time a new version of Python comes out (i.e. #1046).

This diff is actually red (more lines removed than added).

### Commit 2: implements all remaining options
...save for `GIT_OPT_SET_ALLOCATOR`, which I'm pretty confident is useless for Python users, and next to impossible to get right.

[edit] ...and for `GIT_OPT_ADD_SSL_X509_CERT`, which would make pygit2 dependent on OpenSSL since it requires calling OpenSSL APIs directly.

But other options that were missing so far are handled, including three network related that I was really looking forward to:
- `SET_SERVER_TIMEOUT`,
- `SET_SERVER_CONNECT_TIMEOUT`, and
- `ENABLE_HTTP_EXPECT_CONTINUE`.

All new options have a corresponding test.


pygit2/utils.py
```diff
@@ -145,7 +145,7 @@ class StrArray:
 
     __array: 'GitStrrayC | ffi.NULL_TYPE'
     __strings: list['None | ArrayC[char]']
-    __arr: 'ArrayC[char]'
```
ambv, line 148: A strarray is an array of strings, not a single string.

ambv:
I can reproduce the openssl segfault on Ubuntu with:

```
$ ci/cp310/bin/pytest test/test_options.py::test_add_ssl_x509_cert test/test_settings.py::test_ssl_methods

...

test/test_options.py::test_add_ssl_x509_cert PASSED
test/test_settings.py::test_ssl_methods Fatal Python error: Segmentation fault

Current thread 0x0000ffffa870f420 (most recent call first):
  File "/home/ubuntu/pygit2/pygit2/options.py", line 815 in option
  File "/home/ubuntu/pygit2/pygit2/settings.py", line 253 in add_ssl_x509_cert

...
```

So far I had a test with a dummy cert there, which looks like it's harmful. Let me adjust this one test to add an actual cert.

ambv:
It turns out GIT_OPT_ADD_SSL_X509_CERT doesn't accept a byte stream of the cert, but instead expects a pointer to an OpenSSL X509 struct. Since I don't want to make pygit2 directly dependent on OpenSSL, I decided to skip support GIT_OPT_ADD_SSL_X509_CERT in this PR. After all, we can use GIT_OPT_SET_SSL_CERT_LOCATIONS instead.

jdavid:
Thanks!

PR #1410, opened 2025-08-13
Deprecate Remote.ls_remotes in favor of Remote.list_heads

jorio:
This PR implements changes discussed in https://github.com/libgit2/pygit2/pull/1397#issuecomment-3182758075


jdavid:
Thanks @jorio 
I think `ls_remotes` should be fully backwards compatible. I mean, it should return `None` as before instead of the zero Oid (for the `loid` field).

PR #1412, opened 2025-08-19
git_patch_from_diff may return a NULL patch without an error

jorio:
`Diff.__getitem__` used to assume (incorrectly) that when `git_patch_from_diff` returns 0 (success), it also creates a valid `git_patch`.

However, per the [docs for git_patch_from_diff](https://libgit2.org/docs/reference/main/patch/git_patch_from_diff), "For an unchanged file (...) no git_patch will be created, **the output will be set to NULL**".

In practice, `git_patch_from_diff` doesn't always return a NULL patch if the file is unchanged. However, I found a scenario involving CRLF filters that consistently triggers this edge case (success code with NULL patch).

This PR makes `Diff.__getitem__` return `None` in this case. This prevents passing NULL to `wrap_patch`, which would cause undefined behavior in release builds (because [this assertion](https://github.com/libgit2/pygit2/blob/d88fa3dcbab21a7eb33ff354222c4ec8e5ff6ece/src/patch.c#L48) would be bypassed).

### DIFF
--- a/test/test_diff.py
+++ b/test/test_diff.py
@@ -36,6 +36,8 @@ import pygit2
 from pygit2 import Diff, Repository
 from pygit2.enums import DeltaStatus, DiffFlag, DiffOption, DiffStatsFormat, FileMode
 
+from .utils import diff_safeiter
+
 COMMIT_SHA1_1 = '5fe808e8953c12735680c257f56600cb0de44b10'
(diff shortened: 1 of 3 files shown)


jorio:
Hi @jdavid! My latest changes are ready for review. I think pygit2 would benefit from eliminating this undefined behavior (until then, the test scenario can cause a segfault in my git client ;))

PS: I'm not sure why the "Wheels for linux-arm" check failed to start...

PR #1461, opened 2026-04-30
Fix for TypeError in Repository.ident (plus typo fixes)

rapidcow:
As of pygit2 1.19.2 (on my Python 3.14.4, Linux 6.19.14-artix1-1 x86_64), the `Repository.ident` property triggers TypeError:

```
Python 3.14.4 (main, Apr 15 2026, 07:34:47) [GCC 15.2.1 20260209] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygit2; pygit2.__version__
'1.19.2'
>>> repo = pygit2.init_repository('/tmp/foo')
>>> repo.ident
  File "<python-input-2>", line 1, in <module>
    repo.ident
  File "/usr/lib/python3.14/site-packages/pygit2/repository.py", line 1588, in ident
    return (ffi.string(cname).decode('utf-8'), ffi.string(cemail).decode('utf-8'))
            ~~~~~~~~~~^^^^^^^
TypeError: string(): unexpected cdata 'char * *' argument
```

`cname` and `cemail` are `char **` variables, which are pointers to C-strings, not the C-strings themselves. So they should be passed as `cname[0]` and `cemail[0]`.

In addition, `cname[0]` and `cemail[0]` appear to be NULL by default... so clearly we have to handle that separately.

Either way, I wrote up a patch (plus a typo I happened to catch when trying to skim the docs for other mentions of identity). Not sure if this is the right way to contribute here... any feedback\* is appreciated! :)

\*Specifically, (1) I'm not certain if `utils.maybe_string()` is the right tool for this, since the original UTF-8 decode mode was `"strict"`, but is now `"surrogateescape"` (but I don't want to introduce a separate function just for this, so...) Also (2) I used PEP 585 syntax for the tuple annotation, while keeping Optional since it has already been imported (pygit2 itself seems to use a mix of pre- and post-PEP 585 syntax, and ruff doesn't appear to complain either way, so I picked the ones I liked the best without importing more names...)

### DIFF
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -1278,7 +1278,7 @@ class BaseRepository(_Repository):
         Example::
 
             >>> repo = pygit2.Repository('.')
-            >>> repo.stash(repo.default_signature(), 'WIP: stashing')
+            >>> repo.stash(repo.default_signature, 'WIP: stashing')
         """
 
         opts = ffi.new('git_stash_save_options *')
@@ -1347,7 +1347,7 @@ class BaseRepository(_Repository):
         Example::
 
             >>> repo = pygit2.Repository('.')
-            >>> repo.stash(repo.default_signature(), 'WIP: stashing')
+            >>> repo.stash(repo.default_signature, 'WIP: stashing')
             >>> repo.stash_apply(strategy=CheckoutStrategy.ALLOW_CONFLICTS)
         """
         with git_stash_apply_options(
--- a/pygit2/repository.py
+++ b/pygit2/repository.py
@@ -81,7 +81,7 @@ from .references import References
 from .remotes import RemoteCollection
 from .submodules import SubmoduleCollection
 from .transaction import ReferenceTransaction
-from .utils import StrArray, to_bytes
+from .utils import StrArray, maybe_string, to_bytes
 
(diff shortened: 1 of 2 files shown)


rapidcow:
The commit wording was a bit bad, sorry...

```
# git range-diff @{u}~3..@{u} @~3..@
1:  ea5b51b7 ! 1:  d32573ec fix(repository): dereference reflog identity correctly
    @@ Commit message
         as C-strings, when they really are pointers to C-strings.
         Fix by dereferencing the pointer to a C-string first.
     
    -    Furthermore, Repository.ident assumes that libgit2 holds
    -    a valid C-string, when the reflog identity is initialized
    +    Furthermore, Repository.ident assumes that said variables
    +    hold valid C-strings, when the reflog identity is initialized
         to NULL, and can be explicitly set to such with `set_ident
         (None, None)`.  So map None to ffi.NULL (which maybe_string()
         happens to already do, so I simply delegated to that...).
2:  9b65c2d6 = 2:  379b4d82 fix(repository): amend annotations on reflog identity
3:  1deb18ba = 3:  e7bd6d26 docs(repository): fix default_signature typos
```

Issue #1451, opened 2026-03-05
Repository.status() has an unicode issue for badly encoded path


FabriceSalvaire:
To trigger the bug, create a dummy file in a git repository with a (Linux bytes) path set to `'éléphant'.encode('latin1')`
Then call the method `Repository.status()`
It will raise 
```
repository.status()
~~~~~~~~~^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 20: invalid continuation byte
```
This is a general issue with Python3 where we have to explicitly use surrogates...

ralian:
I did some work digging into this over the past weekend. Turns out there were a few other areas where pygit2 wasn't compliant with PEP383 and other encoding recommendations. I added some tests - haven't tested on Windows yet but will do soon.

jdavid:
Should be fixed now, closing.

PR #1469, opened 2026-06-30
encoding: improve UTF-8 decode strategy

ralian:
Work related to #1451

### DIFF
--- a/pygit2/__init__.py
+++ b/pygit2/__init__.py
@@ -365,7 +365,7 @@ from .repository import Repository
 from .settings import Settings
 from .submodules import Submodule
 from .transaction import ReferenceTransaction
-from .utils import to_bytes, to_str
+from .utils import decode_fs_path, encode_fs_path, to_bytes, to_str
 
 # Features
 features = enums.Feature(C.git_libgit2_features())
@@ -430,7 +430,7 @@ def init_repository(
     options.mode = mode
 
     if workdir_path:
-        workdir_path_ref = ffi.new('char []', to_bytes(workdir_path))
+        workdir_path_ref = ffi.new('char []', encode_fs_path(workdir_path))
         options.workdir_path = workdir_path_ref
 
     if description:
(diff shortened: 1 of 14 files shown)


ralian:
I did realize I ran ruff with my settings from another project. I'll have to amend this when I get a chance, but let me know if this change looks ok otherwise.

jdavid:
Thanks @ralian 

PEP-383 says:

>  File names, environment variables, and command line arguments are defined as being character data in POSIX...

In pygit2 we don't have vars or args I think, so we only care about paths.

`maybe_string/to_bytes` are general helpers used not only for filesystem paths. I would rather add 2 new helpers: for example `decode_fs_path/encode_fs_path`; then the work will be to go through the code and use the correct helper (if in doubt keep the old helper call and flag it for discussion).

Also, there is a reference leak in `Repository_status` :

```c
-        err = PyDict_SetItemString(dict, path, status);
+        PyObject *py_path = PyUnicode_DecodeFSDefault(path);
+        if (py_path == NULL)
+            goto error;
+
+        err = PyDict_SetItem(dict, py_path, status);
+        Py_DECREF(py_path);
         Py_CLEAR(status);
```

If `py_path` is `NULL` then `status` is not cleared.

ralian:
Hi David,

Thanks for the feedback. The new helper functions are a good idea, this weekend I will take a crack at that and fix the half broken test too (I have it fixed locally.)

For the reference leak good point. I don't like using goto for this exact reason, but I was trying to match the style of the rest of the codebase. I will think about a better way to handle this, but worst case I can just clear the status in that branch before the goto.

Will

ralian:
I updated and fixed some things with your suggestions @jdavid . I split out that helper function and think I moved everything appropriate over to it; it increases the footprint of the PR, so I am happy to do this either way.

Let me know what you think!

jdavid:
Hi William,

It just remains to:

- Fix the reference leak in Repository_status
- Fix the mypy errors in https://github.com/libgit2/pygit2/actions/runs/29102219007/job/86527997093

ralian:
My apologies, I had run ruff but not mypy. I rewrote that part of repository_status but obviously had the same issue, so I went ahead and just wrapped an extra `Py_CLEAR(status);` in that code path.

ralian:
Once again, forgot to amend with the mypy fixes. Apologies, still getting used to this workflow.

jdavid:
Merged, thanks!

Issue #970, opened 2020-02-03
Access violation accessing repo config entries


GardenTools:
When I try to access a list of the repo config, python terminates abruptly with the message

> Process finished with exit code -1073741819 (0xC0000005)

Example code that causes the issue:

```
import pygit2
repo = pygit2.Repository(path_to_git_repo)
items = list(repo.config)
for item in items:
    print(item.name, item.value)
```
by comparison the following is fine:
```
import pygit2
repo = pygit2.Repository(path_to_git_repo)
items = repo.config
for item in items:
    print(item.name, item.value)
```
and prints out the names and values of the configuration entries would be expected.

It isn't necessary to iterate over the list to cause the problem, the following will result in the same issue
```
items = list(repo.config)
print(items[0].name, items[0].value)
```

pygit2 version: 1.0.2 
python version: 3.7.4 (64-bit)
OS version: Windows 10

---

Doing a little digging https://libgit2.org/libgit2/#HEAD/group/config/git_config_next
states "The pointers returned by this function are valid until the iterator is freed." The ConfigEntry object contains the pointer returned by git_config_next, so it seems these get freed when the iteration is complete, and the ConfigEntry elements of the list are left pointing to invalid memory.

GardenTools:
A workaround is to access the fields of each ConfigEntry and store those whilst the config is being iterated, rather than continuing to reference Config entry after ConfigIterator.\_\_del\_\_ has been called, for example the following works:
```
    configs = {entry.name: entry.value for entry in repo.config}
    for item, val in configs.items():
        print(item, val)
```

beamerblvd:
This is almost certainly a symptom of https://github.com/libgit2/libgit2/issues/7307. It's either a legit libgit2 bug, or pygit2 is improperly using `git_config_entry_free` with iterators when it shouldn't be. It's unclear to me because the libgit2 documentation doesn't say.

PR #1473, opened 2026-07-06
Closes #1472: Implement in-memory config backend for config overrides in context manager

### LINKED ISSUE #1472: Feature Request: Support app-level temporary config overrides using libgit2's backend supprt
When using the `git` CLI, you can temporarily override / add config values with `-c <name>=<value>` (and also `--config-env=<name>=<envvar>`, but that's just a wrapper around `-c` to keep secret values off the command line). The important key to `-c` is that it doesn't persist to a file—the config value lives only so long as that process executes, and then it goes away. Pygit2 has no such support for temporary overrides like this—any changes to a `Config` persist to disk.

libgit2 can support this using config backends. Currently, libgit2 comes with two kinds of config backends—file backends (the standard type) and a read-only in-memory backend constructed from strings. We could, in theory, use the read-only in-memory backend, but its use wouldn't be very Pythonic, and also it's not until the unreleased libgit 1.9.5 that it's possible to construct this backend from the public API (prior to 1.9.5, it was private). 

But libgit2's public API supports implementing and attaching any arbitrary custom backend, and it's fairly straightforward (if not verbose) to implement a mutable in-memory backend with CFFI + Python.

As an aside, pygit2 is also lacking a wrapper around `git_config_open_default`, which is relevant to the approach I'm taking.

With the pull request I'm posting shortly, `Config` (without any public API breakage) becomes a fully-functional base class that, on its own, does what it has always done: wraps access to configuration _files_. Two new classes `DefaultConfig` and `RepositoryConfig` now extend `Config`: the former wrapping `git_config_open_default` and the latter wrapping `git_repository_config`/`git_repository_config_snapshot`. Both of these subclasses can now act as context managers for in-memory config overrides. An in-memory backend attached with level `APP` supports these overrides. When the context manager is entered, the in-memory backend is made the highest priority for writes. When the context manager is exited, the in-memory backend is cleared and made the lowest priority for writes. Within the scope of this context manager, Git operations use these overrides. A hypothetical Python-based alternative Git CLI could, then, implement `-c` and `--config-env` using this context manager.

The detailed documentation I have written for these classes explains in more detail.

beamerblvd:
This implements #1472. It makes several changes to configs in pygit2, and includes substantial documentation improvements to explain the new classes and methods:

- `Config` has several internal changes, but externally, its use and behavior is the same.
- `DefaultConfig` is a new class that extends `Config` and wraps `git_config_open_default`, which previously was not accessible in pygit2.
- `RepositoryConfig` is a new class that extends `Config` and wraps `git_repository_config` / `git_repository_config_snapshot` (previously these were wrapped directly by `Config`).
- When `DefaultConfig` and `RepositoryConfig` are constructed, they construct an in-memory config backend with level `ConfigLevel.APP` that, initially, is unused. When instances of these classes are entered as a context manager, that in-memory backend is given sole write priority. When instances of these classes are exited as a context manager, that in-memory backend in cleared, and the default write priority is restored.
- To address https://github.com/libgit2/libgit2/issues/7307 and #970 (properly), the memory management of iterators and `ConfigEntry` has been refactored.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -209,6 +209,7 @@ Authors:
     Michał Górny
     Mukunda Rao Katta
     Na'aman Hirschfeld
+    Nicholas Williams
     Nicolas Rybowski
     Nicolás Sanguinetti
     Nikita Kartashov
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -2,18 +2,27 @@
 Configuration files
 **********************************************************************
 
-.. autoclass:: pygit2.Repository
-   :members: config
-   :noindex:
+.. autoclass:: pygit2.repository.BaseRepository
+   :members: config, config_snapshot
 
 
-The Config type
+The Config types
 ================
 
 .. autoclass:: pygit2.Config
    :members:
    :undoc-members:
-   :special-members: __contains__, __delitem__, __getitem__, __iter__, __setitem__
+   :special-members: __contains__, __delitem__, __getitem__, __init__, __iter__, __setitem__
+
+.. autoclass:: pygit2.DefaultConfig
+   :members: __enter__, __exit__
+   :undoc-members:
+   :special-members: __init__
+
+.. autoclass:: pygit2.RepositoryConfig
+   :members: __enter__, __exit__
+   :undoc-members:
+   :special-members: __init__
 
 
 The ConfigEntry type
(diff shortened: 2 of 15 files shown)


jdavid:
Thanks @beamerblvd 

This PR is fairly big, would you be willing to split it into smaller PRs? For example:

1. Fix ConfigEntry memory safety / iterator lifetime.
2. Add DefaultConfig and RepositoryConfig.
3. Add the in-memory backend and context-manager support.

That would make each piece much easier to review and merge incrementally.

Issue #1488, opened 2026-09-04
Deadlock in Blob__write_to_queue/git_filter_list_stream_blob


Werkov:
Here's rough code that I have (despite I can reproduce the issue at will, the setup is non-trivial and there are some data dependencies, so I present hopefully the relevant parts).

```python
      # my code
      def from_source_git(self, index, repo, sha, name, current_head, move_upstream):
          tree = repo[sha].peel(pygit2.Tree)
          blob = tree[name]                                                                                                                                             
          bfile = pygit2.BlobIO(blob)
          self._from_file(index, bfile, name, current_head, move_upstream)
          print("loaded")                                                                                                                                                                                                                                                        
          bfile.close()
          print("closed")
...
# pygit2 with some debug prints
class _BlobIO(io.RawIOBase):
...
    def close(self) -> None:                                                                                                            
        try:
            self._ready.wait()
            traceback.print_stack()
            print("gonna wait for %s" % self._writer_closed)                                                                            
            # raise Exception()
            self._writer_closed.wait()
            while self._queue is not None and not self._queue.empty():                                                                  
                self._queue.get()                                                                                                       
            self._thread.join()
        except KeyboardInterrupt:                                                                                                       
            pass
        self._queue = None   
```

So the main thread print this:
```
loaded
  File "/home/mkoutny/tool", line 159, in <module>
    entry.from_source_git(index, source_repo, sha, name, lib.git_sort.oot, True)
  File "/home/mkoutny/lib.py", line 262, in from_source_git
    bfile.close()
  File "/usr/lib64/python3.13/site-packages/pygit2/blob.py", line 99, in close
    traceback.print_stack()
gonna wait for <threading.Event at 0x7fe3f17a3250: unset>
```

AFAICS, it waits for the notification from the writer thread, however, that one is waiting on some locks in queue.put (gdb C stacktrace of the relevant part):

```
#22 0x00007fe3f3093229 in callmethod (tstate=0x55aaa1eeff90, callable=0x7fe37ab30500, format=0x7fe3f25aa56d "y#", va=0x7fe356ffe550) at Objects/call.c:626
#23 PyObject_CallMethod (obj=<optimized out>, name=name@entry=0x7fe3f25aa080 "put", format=format@entry=0x7fe3f25aa56d "y#") at Objects/call.c:645
[...]

PR #1491, opened 2026-09-10
Fix BlobIO cleanup deadlock

Werkov:
This is an attempt at fixing #1488.

It adds a busy wait in the BlobIO cleanup path, it stands the tests and fixes the reported deadlock. (This idea comes from LLM.)

It also adds a testcase with two caveats: there's new binary data in repo (bigrepo.zip) and the test before the fix hangs the whole pytest suite hence it's meant to be run either in isolation (`test/test_blob.py::test_blob_partial_read`) or with the fix. I found no pre-existing test/data repo with large enough file (and the `chunk_size` is not runtime-changeable from tests).


Werkov:
(macos-intel/macos-arm failures (`file_unicode_normalization`) seem unrelated.)

Issue #1004, opened 2020-05-06
Bugs in python backend


aaronschif:
I am creating an issue rather than a PR because I have not been able to create a working Odb or Refdb backend, but I have found several bugs. Please let me know if I am completely misunderstanding this code. 

I believe the issues I cannot get past come from threading in libgit2. I have found that, at times, calling `threading.get_ident` will cause a segfault. I have tried wrapping all C->python code in `PyGILState_Ensure` and `PyGILState_Release`, but this did not work.

I have found the following possible errors:
- https://github.com/libgit2/pygit2/blob/e012bb5bc9e0bdb68ac79f9c12e3f59e3513328a/src/odb_backend.c#L244 Calls `exists_prefix` not `refresh`.
- https://github.com/libgit2/pygit2/blob/e012bb5bc9e0bdb68ac79f9c12e3f59e3513328a/src/refdb_backend.c#L134 `glob` may be null
- https://github.com/libgit2/pygit2/blob/e012bb5bc9e0bdb68ac79f9c12e3f59e3513328a/src/refdb_backend.c#L134 `git_reference* _ref` will double free
- https://github.com/libgit2/pygit2/blob/e012bb5bc9e0bdb68ac79f9c12e3f59e3513328a/src/refdb_backend.c#L409 `PyIter_Check` should be a check for `__iter__` not if the object is currently an iterator.

In my experience, problems like this mean that I am misusing something. Let me know if these appear to be real bugs and I will see if I can at least create test cases to expose them. I don't seem to have a good grasp on the threading segfaults though.

jdavid:
I think you're right about at least some of the bugs, ping @ddevault since he's the author of this code. For the 3rd one I think the link is not correct, should not be https://github.com/libgit2/pygit2/blob/e012bb5bc9e0bdb68ac79f9c12e3f59e3513328a/src/refdb_backend.c#L205 ?

Will give a look at the threading issue.

ddevault:
I think you would be completely bonkers to introduce multithreading into a Python program, let alone a Python program with C modules.

Can you turn your linewise feedback into a patch, for easier commentary and a better path towards merge?

aaronschif:
Threads are common enough in python, especially in web. Either way, my python program does not introduce threads (yet, it is web). I am not familiar with libgit2 internals or python c modules but for the reasons I mentioned earlier, my thought is that threading in libgit is not playing well with python.

Thanks for getting back to me. I will clean up my code create a branch and we can continue.

[...]

PR #1496, opened 2026-09-19
OdbBackend: acquire the GIL in the callbacks, fix write_cb's type argument

cboos:
With an `OdbBackend` implemented in Python, `Index.add(path)` crashes the interpreter (access violation on Windows, Python 3.13, pygit2 1.20.1), single-threaded.

**Cause:** the odb backend callbacks in `src/odb_backend.c` call into Python without holding the GIL. That works when libgit2 is entered from the C extension, which keeps the GIL, but cffi releases it around every C call, so `git_index_add_bypath` reaches the callbacks without it.

**Fix:** `PyGILState_Ensure`/`PyGILState_Release` in each callback, the same pattern as in `src/filter.c`. It is re-entrant, so it is fine when the caller already holds the GIL.

**Second commit:** `write_cb` received its `type` argument through the `n` format (`Py_ssize_t`) though `git_object_t` is an int; on 64-bit Windows it got e.g. 2753074036739 instead of 3 for a blob. It is now passed as `i`.

**Tests:** `test_index_add` (run in a subprocess, since the failure is a crash) and `test_write_cb`. Both fail before and pass after, on Windows x64 with MSVC, CPython 3.13 and libgit2 1.9.7; the full suite passes with the first commit.

Not done here: `src/refdb_backend.c` has the same missing GIL. Possibly related: #1004.

---
This PR (analysis, patch, tests and this description) was prepared by Claude Fable 5.1, working for @cboos.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cboos:
Thanks!

